PR #3, opened 2020-01-24
RFC-0001: Add method __torch_function__ RFC.

hameerabbasi:


ezyang:
I would have expected a discussion of backwards compatibility on this proposal, both with `__torch_function__` as we currently landed it in PyTorch master, and also in comparison to `__array_function__` in Numpy. If we actually change `__torch_function__` to operate on methods, does this break users who expected it to work the same way as in Numpy? If it does break users, is this a fair price to pay? There is no analysis or discussion on this right now.

RFC 0001.md
```diff
@@ -0,0 +1,105 @@
...
+
+s3 = torch.Tensor(1) + t
+isinstance(s3, SubTensor)  # True
+s3.a  # 1
+```
+
+Additionally, it will provide subclass authors hooks to run whenever methods or operators are called, and to modify the result to their specific use-case, perform logging, or otherwise change the result or the action of the method.
+
+## Detailed Description
+We propose the following signature change to `__torch_funcion__`, to make it match NumPy: [[4]]
```
ezyang: function

RFC 0001.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+Additionally, it will provide subclass authors hooks to run whenever methods or operators are called, and to modify the result to their specific use-case, perform logging, or otherwise change the result or the action of the method.
+
+## Detailed Description
+We propose the following signature change to `__torch_funcion__`, to make it match NumPy: [[4]] 
+
+```python
+class SubTensor(torch.Tensor):
+    def __torch_tensor__(self, func, types, args, kwargs):
```
ezyang: You don't say what the old type of `__torch_tensor__` was so I can't tell what the difference is.

@ngoldbaum do you remember why we didn't line up the type exactly with Numpy's type in the beginning?

rgommers: IIRC that was me. Not 100% sure, but I believe I removed (or never added) `types` because it was extra complexity to handle subclasses, and `Tensor` subclasses aren't really a thing today.

ngoldbaum: No reason afaik. That API got hammered out before I started working on the feature so I don’t know why the API is different. It wouldn’t be terribly hard to add `types` to the signature if we wanted to do that.

ngoldbaum:
(although doing so would break any downstream users who are out there)

RFC 0001.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+The reason for this change is necessitated by the need for `super()`. If we set a requirement for `super().__array_function__` to work properly, we would need to provide an easy way for users to signal to `__array_function__` that they are calling to the next-specific implementation. The way we propose to handle this is the same as it is handled in NumPy, albiet not in the context of overriding methods, but rather, in the context of subclasses of `numpy.ndarray` or other classes that implement `__array_function__`.
+
+To access super, one would do the following:
+```python
+class SubTensor(torch.Tensor):
+    def __torch_tensor__(self, func, types, args, kwargs):
+        # Pre-processing here
+        val = super().__torch_function__(func, tuple(t for t in types if not issubclass(t, SubTensor), args, kwargs)
```
ezyang: Missing a closing paren for `tuple(`

RFC 0001.md
```diff
@@ -0,0 +1,105 @@
...
+        # Post processing here
+```
+
+This way `__torch_function__` knows the list of types to dispatch to, and it will _not_ dispatch to `SubTensor` anymore in this example.
+
+We will also recommend that all `Tensor` subclasses make their own methods go through `__torch_function__` via a decorator `@torch_function_dispatch`. However, this will come with a disclaimer: They _must_ accept that their methods are subject to the same processing as any other `torch.Tensor` methods, namely, that all the processing _will necessarily go through `__torch_function__`, even if through superclasses first_.
```
ezyang: NB: `torch_function_dispatch` is dead in master because it was too slow. So this should be taken into account in planning here.

RFC 0001.md
```diff
@@ -0,0 +1,111 @@
...
+        # Defer to internal implementation
+        ret = func._implementation(*args, **kwargs)
+        if type(self) is not Tensor:
+            ret = Tensor.make_subclass(ret, type(self))
+        return ret
+```
+
+This method matches `torch` dispatch rules, so for the most part it's possible to pretend it doesn't exist. This also has the side-effect of passing subclasses through methods, and operators (since all operators are methods).
+
+This corresponds exactly to the implmentation `numpy.ndarray` gains in [[4]], except for the fact that subclasses are passed through via another internal mechanism there, as well as the fact that we are checking subclassing against `type(self)` instead of `Tensor`. This has the side-effect of ensuring unrelated class trees are not merged, which is an inconsistency in NumPy's own design.
```
ezyang: implementation

ezyang:
Procedural note: let's put a more descriptive filename on RFCs, so we can easily tell what's what :)

RFC 0001.md
ezyang: what's "another internal mechanism" there?

RFC 0001.md
```diff
@@ -0,0 +1,111 @@
...
+        val = super().__torch_function__(func, tuple(t for t in types if not issubclass(t, SubTensor)), args, kwargs)
+        # Post processing here
+```
+
+This way `__torch_function__` knows the list of types to dispatch to, and it will _not_ dispatch to `SubTensor` anymore in this example.
+
+We will also recommend that all `Tensor` subclasses make their own methods go through `__torch_function__` via a decorator `@torch_function_dispatch`. However, this will come with a disclaimer: They _must_ accept that their methods are subject to the same processing as any other `torch.Tensor` methods, namely, that all the processing _will necessarily go through `__torch_function__`, even if through superclasses first_.
+
+### Making `torch.Tensor._make_subclass` public API
+`torch.Tensor._make_subclass` will be renamed to `torch.Tensor.make_subclass` and it will become public API.
```
ezyang: Does `make_subclass` correspond to anything in Numpy? Can you write down what the "public API docs" of `make_subclass` would be here?

RFC 0001.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Additionally, it will provide subclass authors hooks to run whenever methods or operators are called, and to modify the result to their specific use-case, perform logging, or otherwise change the result or the action of the method.
+
+## Backwards Compatibility
+### With PyTorch `master` as of writing
+PyTorch `master` pointed to commit hash `957a07ffbd13d8a805f4d718e0282efc5d2bff85` at the time of writing. Any classes implementing `__torch_function__` based on the usage in this commit hash will break completely, due to the differing signature of the protocol. However, as a release hasn't been made with `__torch_function__` in it, this is a minor-impact issue.
+
+### With NumPy
+As we are using a different protocol compared to NumPy `__torch_function__` vs `__array_function__`, there is no difference to the usage for those using NumPy. We propose to delay the issue of allowing the usage of Torch tensors with NumPy functions to a separate RFC.
```
ezyang: Sorry, I wasn't clear in my earlier comment. My question is not allowing PyTorch tensors be used with Numpy functions (although this is an interesting question to pose), but if I am a Numpy user who is familiar with the `__array_function__` API, and I come to PyTorch expecting `__torch_function__` to work the same way, will my expectations be surprised in any way?

RFC 0001.md
ezyang: I think it's safe to say we should do this sooner rather than later, even if other parts of this RFC change in their design.

RFC 0001.md
```diff
@@ -0,0 +1,111 @@
...
+## Detailed Description
+We propose the following signature change to `__torch_function__`, to make it match NumPy: [[4]] 
+
+```python
+class SubTensor(torch.Tensor):
+    def __torch_tensor__(self, func, types, args, kwargs):
+        # Implementation here
+```
+
+The reason for this change is necessitated by the need for `super()`. If we set a requirement for `super().__array_function__` to work properly, we would need to provide an easy way for users to signal to `__array_function__` that they are calling to the next-specific implementation. The way we propose to handle this is the same as it is handled in NumPy, albiet not in the context of overriding methods, but rather, in the context of subclasses of `numpy.ndarray` or other classes that implement `__array_function__`.
```
ezyang: albeit

RFC 0001 — `__torch_function__` for methods of the `torch.Tensor` class.md
```diff
@@ -0,0 +1,115 @@
...
+        # Pre-processing here
+        val = super().__torch_function__(func, tuple(t for t in types if not issubclass(t, SubTensor)), args, kwargs)
+        # Post processing here
+```
+
+This way `__torch_function__` knows the list of types to dispatch to, and it will _not_ dispatch to `SubTensor` anymore in this example.
+
+We will also recommend that all `Tensor` subclasses make their own methods go through `__torch_function__` via a decorator `@torch_function_dispatch`. This decorator was added and then removed for performance reasons, however it will be added back to allow external libraries to interface with the protocol.
+
+However, this will come with a disclaimer: They _must_ accept that their methods are subject to the same processing as any other `torch.Tensor` methods, namely, that all the processing _will necessarily go through `__torch_function__`, even if through superclasses first_.
```
ezyang: Can you elaborate this paragraph. It is not very clear what you mean here. Examples would help.

ezyang:
Umm, the new filename is better, but can we avoid putting hard to quote characters in the filename? :> Preferably no spaces either.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,117 @@
...
+
+We do not propose automatic marking of functions with this decorator due to the potential backwards-compatibility break it could cause, as well as the parameters that are needed in order to allow this to happen (namely the dispatcher, which isn't in our control).
+
+### Making `torch.Tensor._make_subclass` public API
+`torch.Tensor._make_subclass` will be renamed to `torch.Tensor.make_subclass` and it will become public API. This will allow `torch.Tensor.make_subclass` correspond to `numpy.ndarray.view`, with the difference that the latter can also handle viewing as a different `dtype` and not just as a different subclass.
+
+The reason for not choosing the name `view` is that it exists on `torch.Tensor` in an unrelated context. The semantics of this function will be the same as creating a shallow copy of the object and then changing its `__class__`.
```
ezyang: Shallow copy as defined by https://docs.python.org/2/library/copy.html ?

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,117 @@
...
+s2 = t + torch.Tensor(1)
+isinstance(s2, SubTensor)  # True
+s2.a  # 1
+
+s3 = torch.Tensor(1) + t
+isinstance(s3, SubTensor)  # True
+s3.a  # 1
+```
+
+Additionally, it will provide subclass authors hooks to run whenever methods or operators are called, and to modify the result to their specific use-case, perform logging, or otherwise change the result or the action of the method.
```
ezyang: The usage descriptions here seem insufficient. In this paragraph you describe a bunch of other things users could do, but no code!

RFC 0001.md
shoyer: As I wrote over in https://github.com/pytorch/pytorch/pull/30730#issuecomment-583262431, my motivation for the `types` argument in NumPy's `__array_function__` was never about subclasses (which I agree are generally awful). My concern was making it as easy and idiomatic as possible for implementers of `__array_function__` to defer to unrecognized types that might also implement an operation. This is generally a good thing for the ecosystem, but people writing special methods tend not to bother, at least for builtin numeric protocols like `__add__` :).

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+        return super().__torch_function__(
+            func,
+            tuple(
+                t for t in types if not issubclass(t, LoggingTensor)
+            ),
+            args,
+            kwargs
+        )
+```
+
```
rgommers: A discussion of performance is still missing. It should state the requirements:
- no impact on `Tensor` performance
- no deterioration of current `__torch_function__` overhead
- small (sub-us) overhead on `Tensor` subclasses
And whether that's achievable.

Here would be a good place (it falls under "Impact").

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+            ),
+            args,
+            kwargs
+        )
+        # Post processing here
+```
+
+This way `__torch_function__` knows the list of types to dispatch to, and it will _not_ dispatch to `SubTensor` anymore in this example.
+
+We will also recommend that all `Tensor` subclasses make their own methods go through `__torch_function__` via a decorator `@torch_function_dispatch`. This decorator was added and then removed for performance reasons, however it will be added back to allow external libraries to interface with the protocol. It will take a single argument: a dispatcher, i.e. a callable that returns an iterable of all the "duck-Tensors", or possible candidates for classes that may implement `__torch_function__`.
```
rgommers: That `torch_function_dispatch` addition should be added as a main point in your proposal (point 5 in Motivation and Scope section). Especially because you not only want to add it back, but to make it public (NumPy hasn't even done that yet....).

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+
+If a library forgets to add the aforementioned decorator, then the method will no longer dispatch at all to any form of `__torch_function__`. In other words, it will lose support for the protocol.
```
rgommers: Most users won't care about losing support for the protocol. However, I think the (unstated) implication is that the returned tensor is a `Tensor` instance rather than a subclass instance. This should be stated clearly here.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+
+However, this will come with a disclaimer: They _must_ accept that their methods are subject to the same processing as any other `torch.Tensor` methods, namely, that all the processing _will necessarily go through `__torch_function__`, even if through superclasses first_. Specifically, processing may pass through a superclass's `__torch_function__` implementation before coming back to a subclass's internal implementation. A workaround would be to define a `__torch_function__` on the same class that directly calls the internal implementation, or follows the same pattern as the example `__torch_function__` defined on `torch.Tensor`. This will allow subclasses to behave exactly like `torch.Tensor`, while still allowing access to method overriding via the `__torch_function__` protocol.
```
rgommers: This paragraph needs a rewrite. It's unclear whether it's important (it's just a disclaimer), it's unclear what a user should actually do (does "must accept ..." imply they need to add some code or not?), and what the penalty is.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+        # Post-processing
+        return ret
+```
+
+In this instance, processing would follow the `__torch_function__` protocol. This means that control would end up in `SubTensor.__torch_function__`, go to `Tensor.__torch_function__` from there and and then come to `SubSubTensor.__add__`, from where it would go to `Tensor.__add__`, and then back up the stack in the reverse order. This means that great care needs to be taken when writing `SubTensor.__torch_function__` to take into account the fact that it has to handle subclass methods.
+
+We do not propose automatic marking of functions with this decorator due to the potential backwards-compatibility break it could cause, as well as the parameters that are needed in order to allow this to happen (namely the dispatcher, which isn't in our control).
+
+### Making `torch.Tensor._make_subclass` public API
+`torch.Tensor._make_subclass` will be renamed to `torch.Tensor.make_subclass` and it will become public API. This will allow `torch.Tensor.make_subclass` correspond to `numpy.ndarray.view`, with the difference that the latter can also handle viewing as a different `dtype` and not just as a different subclass.
```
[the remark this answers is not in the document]
rgommers: Also, no actual motivation is given for why `make_subclass` needs to be public. How about adding both a sentence on that and a usage example?

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+
+The reason for not choosing the name `view` is that it exists on `torch.Tensor` in an unrelated context. The semantics of this function will be the same as creating a shallow copy of the object and then changing its `__class__`, which means that the underlying data pointer will remain the same.
+
+### Generic implementation of `__torch_function__`
+`torch.Tensor` will gain a generic `__torch_function__` of the following form:
```
rgommers: Start the sentence with `"In order to .... <reason for doing this>"`

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+
+This corresponds exactly to the implementation `numpy.ndarray` gains in [[4]], except for the fact that subclasses are passed through via another internal mechanism (namely the `__array_finalize__` protocol) there, as well as the fact that we are checking subclassing against `type(self)` instead of `Tensor`. This has the side-effect of ensuring unrelated class trees are not merged, which is an inconsistency in NumPy's own design. Specifically, consider the example of two direct subclasses of `torch.Tensor`. Both will return `NotImplemented`, and therefore, the check will fail and `TypeError` will be raised.
+
+Since subclasses are checked before superclasses in `__torch_function__`, it is guaranteed that the subclass implementation will be called first. In this instance, since `type(self)` is a subclass of all types, the code will continue. Since `type(self)` is not `torch.Tensor`, a view into the original data is created and returned.
+
+This also works for all operators: `__add__`, `__getitem__` and so on since in Python these operators are just dunder methods of the corresponding class.
+
```
rgommers: Can you add an `Implementation` section? No need for a lot of research to fill it, just state what you know now. We spent a ton of time on the C++ implementation of the dispatching, so a hint as to whether it can simply be mirrored in the method version or not would be nice (perhaps @ngoldbaum has a sense here about how parallel method/function code is).

RFC-0001-torch-function-for-methods.md
hameerabbasi: >     * no impact on `Tensor` performance

This seems undoable, from what I understand, since `Tensor` methods that don't take arguments don't go through the processing of `Tensor` arguments, and they'll have to. In addition, for those that _do_ take arguments, `self` will have to be added in the list of processed arguments, so there will be a minor hit. This is regardless of the approach we take (`__torch_finalize__` will also have these drawbacks).

RFC-0001-torch-function-for-methods.md
hameerabbasi: Added.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+PyTorch `master` pointed to commit hash `957a07ffbd13d8a805f4d718e0282efc5d2bff85` at the time of writing. Any classes implementing `__torch_function__` based on the usage in this commit hash will break completely, due to the differing signature of the protocol. However, as a release hasn't been made with `__torch_function__` in it, this is a minor-impact issue. This brings the design of `__torch_function__` more in line with NumPy's `__array_function__`, and one familiar with NumPy's protocol could transition to PyTorch's take on it without too many surprises, with the caveat that it could also receive methods rather than functions.
+
+### With NumPy
+As we are using a different protocol compared to NumPy `__torch_function__` vs `__array_function__`, there is no difference to the usage for those using NumPy. We propose to delay the issue of allowing the usage of Torch tensors with NumPy functions to a separate RFC. 
+
+## Detailed Description
+We propose the following signature change to `__torch_function__`, to make it match NumPy: [[4]]
+
+```python
+class SubTensor(torch.Tensor):
```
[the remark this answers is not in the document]
hameerabbasi: I restructured things and moved them around: See the section on `super`.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+
+This corresponds exactly to the implementation `numpy.ndarray` gains in [[4]], except for the fact that subclasses are passed through via another internal mechanism (namely the `__array_finalize__` protocol) there, as well as the fact that we are checking subclassing against `type(self)` instead of `Tensor`. This has the side-effect of ensuring unrelated class trees are not merged, which is an inconsistency in NumPy's own design. Specifically, consider the example of two direct subclasses of `torch.Tensor`. Both will return `NotImplemented`, and therefore, the check will fail and `TypeError` will be raised.
+
+Since subclasses are checked before superclasses in `__torch_function__`, it is guaranteed that the subclass implementation will be called first. In this instance, since `type(self)` is a subclass of all types, the code will continue. Since `type(self)` is not `torch.Tensor`, a view into the original data is created and returned.
+
+This also works for all operators: `__add__`, `__getitem__` and so on since in Python these operators are just dunder methods of the corresponding class.
+
+
```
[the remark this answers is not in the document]
hameerabbasi: Added.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,177 @@
...
+Additionally, it will provide subclass authors hooks to run whenever methods or operators are called, and to modify the result to their specific use-case, perform logging, or otherwise change the result or the action of the method. For example:
+
+```python
+import logging
+
+class LoggingTensor(torch.Tensor):
+    def __torch_function__(self, func, types, args, kwargs):
+        logging.info(f"func: {func!r}, args: {args!r}, kwargs: {kwargs!r}")
+        return super().__torch_function__(
```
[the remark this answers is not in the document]
hameerabbasi: Added.

RFC-0001-torch-function-for-methods.md
rgommers: Hmm, that seems important to estimate first, in a bit more detail than I did in gh-22402, since it could be a showstopper.

> since `Tensor` methods that don't take arguments don't go through the processing of Tensor arguments, and they'll have to

would be good to know which these are

> In addition, for those that _do_ take arguments, `self` will have to be added in the list of processed arguments, so there will be a minor hit.

An estimate for this impact would be good. Should be straightforward, by comparing overhead on current `__torch_function__` for functions with different number of arguments. Maybe that info is already in one of @ngoldbaum's PRs.

RFC 0001.md
rgommers: I don't think there are more than a few handful of users that have built up an intuition here since it's so new, nor do I think including methods will be particularly surprising.

RFC 0001.md
rgommers: > was never about subclasses (which I agree are generally awful)

I kind of disagree. `numpy.ndarray` subclasses give us a bad taste in the mouth because of the history of `numpy.matrix` and other badly written subclasses. But in principle they make sense. Also for PyTorch there's a lot of interest, and the use cases are solid - and letting those users write a whole `torch.Tensor`-like object is definitely not feasible in most cases.

RFC-0001-torch-function-for-methods.md
hameerabbasi: > would be good to know which these are

These would be methods like `sum`, `prod`, and others that don't take a `Tensor`-like other than `self`.

hameerabbasi: > An estimate for this impact would be good. Should be straightforward, by comparing overhead on current `__torch_function__` for functions with different number of arguments. Maybe that info is already in one of @ngoldbaum's PRs.

The problem with this approach is it takes into account the protocol overhead along with the checking for `Tensor` operands. In any case, I took a look at https://github.com/pytorch/pytorch/pull/27064#issuecomment-554601189 (@ngoldbaum, please correct me if there are more recent benchmarks), but the benchmarks are too all over the place to come up with an accurate number (the issue @peterbell10 mentioned). I'd ballpark the upper bound of performance hit to be in the range 0.1 to 1 microsecond, for `Tensor`-only operands.

rgommers: That's not the latest PR you're referencing: https://github.com/pytorch/pytorch/pull/32194 was the final version with inlining dispatch rather than using the decorator; overhead is <100ns (see PR description).

hameerabbasi: Well, that's Python, but it doesn't have comparisons on input argument size. In any case, the overhead should be sub-50 ns.

rgommers: Okay, sounds sensible for parsing one more parameter. If you add a sentence on that to the new `Performance` section as the estimate for parsing `self`, then I think that's enough for now.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,332 @@
...
+## Motivation and Scope
+Quoting [[1]], [[2]] and [[3]], the goals of this proposal are:
+
+1. Support subclassing `torch.Tensor` in Python
+2. Preserve `Tensor` subclasses when calling `torch` functions on them
+3. Use the PyTorch API with `torch.Tensor`-like objects that are _not_ `Tensor`
+   subclasses
+4. Preserve `Tensor` subclasses when calling `Tensor` methods.
+5. The ability to give external libraries a way to also define
+   functions/methods that follow the `__torch_function__` protocol.
```
rgommers: This one is unrelated, it's kind of a by-product coming from the implementation. It's not a goal, and TBD if actually desired; I'd be inclined to hold off on this.

So that's a separate decision to make. Let's move it out of this list and further down. I'd say expose `torch_function_dispatch` if needed as a tool for subclassing, but let's document for now that it is advised not to be used for anything but subclass methods for now.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,332 @@
...
+6. Propagating subclass instances correctly also with operators, using
+   views/slices/indexing/etc.
+7. Preserve subclass attributes when using methods or views/slices/indexing.
+8. A way to insert code that operates on both functions and methods uniformly
```
rgommers: This one is also separate from subclassing, so would also mark it separately as such. It is an oft-requested feature though. So there's 6 subclassing-related goals, then this goal, and then a maybe-goal.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,332 @@
...
+Achieving interoperability with NumPy and adopting its array protocols is out
+of scope for this proposal and we propose to defer it to a later proposal.
+
+We propose to solve this problem with the following changes to PyTorch:
+
+1. Make methods and operators of `torch.Tensor` go through the
+   `__torch_function__` machinery.
+2. Add a `types` argument to `__torch_function__`, to make it match NumPy's
+   `__array_function__`.
+3. Make `torch.Tensor._make_subclass` public API.
```
rgommers: Rephase suggestion: `Make `torch.Tensor._make_subclass` public API (as `make_subclass`).

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,332 @@
...
+
+To make the protocol operate only on functions rather than methods, one can
+check for `func not in type(self).__dict__.values()`. To check for operators
+and/or indexing, one can check `func.__name__.endswith("__")`.
+
+### Performance
+There are a few requirements for the performance of this proposal, when
+implemented:
+
+* No sub-100 ns deterioration per function call on `torch.Tensor` objects.
```
rgommers: This is phrased wrong. "No sub-100 ns" says more than 100 ns is okay. Propose this change:

1. Requirement is: "no deterioration for performance of `torch.Tensor` instances."
2. Under the requirements list say: "The first requirement cannot be fully met. Due to the need to add `self` as an extra parameter in argument parsing, we expect a small (sub 100 ns) impact for method calls (and no impact for function calls)".

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,332 @@
...
+data is created and returned.
+
+This also works for all operators: `__add__`, `__getitem__` and so on since in
+Python these operators are just dunder methods of the corresponding class.
+
+### The need for `super().__torch_function__`
+To access super, one would do the following:
+```python
+class SubTensor(torch.Tensor):
+    def __torch_tensor__(self, func, types, args, kwargs):
```
rgommers: still a wrong method name here, should be `__torch_function__`

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,332 @@
...
+equivalent to the code above, if the `Tensor`-likes are all `Tensor`, the
+internal implementation is called immediately. This is done as a performance
+optimisation to avoid overhead for concrete `Tensor` objects.
+
+### Generic implementation of `__torch_function__`
+`torch.Tensor` will gain a generic `__torch_function__` of the following form:
+
+```python
+class Tensor:
+    def __torch_tensor__(self, func, types, args, kwargs):
```
rgommers: also wrong method name

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,332 @@
...
+parameters that are needed in order to allow this to happen (namely the
+dispatcher, which isn't in our control).
+
+### Making `torch.Tensor._make_subclass` public API
+`torch.Tensor._make_subclass` will be renamed to `torch.Tensor.make_subclass`
+and it will become public API. This method will create an object that has the
+same data pointer as the original object, which means that modifications to
+this will be reflected in the original object. More or less, it will have the
+same effect as modifying an object's `__class__` attribute in Python.
+
```
rgommers: This still needs a motivation, it's not explained anywhere I think. Is it _only_ for use within `SubTensor.__torch_function__` as in the one example, or for other reasons too? Please give the 1-2 sentence motivation, and perhaps a concrete example of how to use it if there's another reason.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,343 @@
...
+We propose to solve this problem with the following changes to PyTorch:
+
+1. Make methods and operators of `torch.Tensor` go through the
+   `__torch_function__` machinery.
+2. Add a `types` argument to `__torch_function__`, to make it match NumPy's
+   `__array_function__`.
+3. Make `torch.Tensor._make_subclass` public API by renaming it to `torch.Tensor.make_subclass`.
+4. Make `torch.Tensor` gain a generic implementation of `__torch_function__`.
+
+## Usage and Impact
```
ezyang: Thank you, these examples are great!

RFC-0001-torch-function-for-methods.md
hameerabbasi: Added just below this comment.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,343 @@
...
+```python
+import logging
+
+class LoggingTensor(torch.Tensor):
+    def __torch_function__(self, func, types, args, kwargs):
+        logging.info(f"func: {func.__name__}, args: {args!r}, kwargs: {kwargs!r}")
+        return super().__torch_function__(
+            func,
+            tuple(
+                t for t in types if not issubclass(t, LoggingTensor)
```
ezyang: It seems annoying that all users have to write this boilerplate whenever they make parent calls.

ezyang: BTW, I don't think we need to design for this here, but I am wondering: do you know if this pattern works correctly for multiple inheritance in Python?

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,343 @@
...
+implemented:
+
+1. No deterioration for function/method calls on `torch.Tensor` objects.
+2. No deterioration of current `__torch_function__` overhead
+3. Sub-µs impact on the performance of subclasses not implementing
+  `__torch_function__`.
+
+Requirement 1 seems unachievable due to the structure of the code at this
+point, as processing will have to be added for `self` in methods, as well as
+adding argument processing for methods where it doesn't already exist. We think
```
ezyang: Err, assuming that we have `__torch_function__` existing overhead, shouldn't there be no *further* overhead here? (Are you saying that the overhead from functions will then propagate to methods? We agreed that this overhead was fine for the `__torch_function__` patch.)

RFC-0001-torch-function-for-methods.md
hameerabbasi: I think we could provide an `exclude_type` helper function, but beyond that, I'm not sure what we can do.

RFC-0001-torch-function-for-methods.md
hameerabbasi: Okay, there are two issues here:

1. The code that extracts out `Tensor`-likes for `__torch_function__` handling doesn't process `self`, and processing needs to be added for it. That will incur a minor cost.
2. For methods that only take `self` as an argument and no other `Tensor`-like, and are implemented in C++, those are skipped from the processing entirely, and processing needs to be added for them. That will incur a cost for these methods. These are methods like `tensor.sum()`.

hameerabbasi: A third issue applies to methods defined in Python, to these the overhead will propagate.

hameerabbasi: Added this to the proposal.

RFC-0001-torch-function-for-methods.md
hameerabbasi: And it will, you can just add `type(self).__bases__[base_index_you_want_here]` to the list of `types`, and assuming it implements `__torch_function__` and is hit in the MRO, it'll work fine.

ezyang:
Sorry about the review delay, I have a fairly complicated line of thought forming but I need to line it up with what the proposal says right now, and I'm not too sure how to express it right now.

ezyang:
So, the good news: I think that, as far as the core fundamentals go, this proposal is in a good spot. What I think is left is clearer communication in the RFC, and some hashing out of details.

On the communication front, a big chunk of what I think is missing from this PR is something like this diagram:

image

This is diagram is based on my current reading of your RFC, and some extrapolation of unstated points in the RFC. A few notable points:

* In this diagram, I have given up on the idea that you can affect the "PyTorch dispatch mechanism" (the right side of the diagram, the one involved with `__torch_function__`) by directly specifying methods on the class. So I imagine that processing of tensors happens in two phases, one where we do traditional Python style OOP, passing down the tensor until we get to a base method, and then a separate dispatch chain where we call the user defined `__torch_function__` global overload, or defer to the built-in backend. If a user what's to define a "special" case for the PyTorch dispatcher, they should do this by adding an if statement to their `__torch_function__` implementation (they can use @ngoldbaum's idiom of a hash table to conveniently define these). This might be a little confusing, but users can avoid the confusing case by sticking to only methods, or only `__torch_function__`, and not mixing both unless they know what they're doing.
* Therefore, I *disagree* with the use of `torch_function_dispatch` on `__add__` in Section "The need for `super().__torch_function__`". You discuss a situation where `__torch_function__` would call `__add__`. I infer from your discussion in Section "Protocol support for external libraries" that this invocation could occur because a `torch_function_dispatch` wrapped method will invoke `__torch_function__` set to the original method that was wrapped by `torch_function_dispatch` (the concrete implementation of `__add__` as specified in this example). This certainly "works" but I think it will do something very unexpected to the user. Here is the sequence of events that I predict will happen:

  image

  An easier way to understand what is going on is to simply imagine that instead of naming the decorated function `__add__`, we name it `my_special_add` (and have it call `super().__add__()`. Now you can see that this is simply an external library method (as described in "Protocol support for external libraries") and its "backend" implementation involves invoking `__add__`  on the superclass (aka `Tensor.__add__`), which itself involves another go around the `__torch_function__` mechanism. I suspect this will all work, and maybe even some poor user will find a good use for it, but we should certainly not recommend that things be used this way.
* Therefore, I think that the `issubclass` test, and the `types` parameter in general, are not necessary. I will go through the RFC and comment places where I think it can be removed.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,358 @@
...
+class LoggingTensor(torch.Tensor):
+    def __torch_function__(self, func, types, args, kwargs):
+        logging.info(f"func: {func.__name__}, args: {args!r}, kwargs: {kwargs!r}")
+        return super().__torch_function__(
+            func,
+            tuple(
+                t if not issubclass(t, LoggingTensor) else torch.Tensor for t in types
+            ),
+            args,
+            kwargs
```
ezyang: I think there is a teaching moment here. Why should we call `super().__torch_function__` rather than directly calling `func` which is the backend implementation? Answer: if we called func directly, we wouldn't wrap the resulting tensor in a LoggingTensor (thus losing the Logging-ness for downstream calls). This could be a valid choice for some kinds of tensors! But not for this one, probably.

ezyang: Actually, it seems like calling `func` would not work at all, if we follow the calling convention from existing `__torch_function__`. The existing calling convention is that `func` is exactly something like `torch.Tensor.__add__`, which is a *dispatching* function; if you called `func` directly you would infinite loop, if `func` is a function that goes to the dispatcher, which would in turn call `__torch_function__` again (infinite loop). This kind of suggests we need either (1) a way to name functions that bypasses dispatch, or (2) a way to invoke a function without triggering dispatch. I'm guessing this is what the `_implementation` field on these functions does; worth calling this out explicitly in the RFC!

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,358 @@
...
+```python
+class Tensor:
+    def __torch_function__(self, func, types, args, kwargs):
+        if not all(issubclass(type(self), t) for t in types):
+            return NotImplemented
+        
+        # Defer to internal implementation
+        ret = func._implementation(*args, **kwargs)
+        if type(self) is not Tensor and isinstance(ret, Tensor):
+            ret = Tensor.make_subclass(ret, type(self))
```
ezyang: Does this work for non-methods, e.g., `torch.function`?

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,358 @@
...
+of extracting an iterable of objects that *could* be `Tensor`-like.
+
+### Generic implementation of `__torch_function__`
+`torch.Tensor` will gain a generic `__torch_function__` of the following form:
+
+```python
+class Tensor:
+    def __torch_function__(self, func, types, args, kwargs):
+        if not all(issubclass(type(self), t) for t in types):
+            return NotImplemented
```
ezyang: If we eliminate `types` as an explicitly passed in argument, the modified test would simply extra tensor-likes from args and then check that the types are consistent. We do this already to determine what `__torch_function__` to call, so I think for the internal implementation this would not be a big deal to do. I don't *think* that users need to do this, but I'm willing to be convinced otherwise.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,358 @@
...
+                t for t in types if not issubclass(t, SubTensor)
+            ),
+            args,
+            kwargs
+        )
+        # Post processing here
+```
+
+This way `__torch_function__` knows the list of types to dispatch to, and it
+will _not_ dispatch to `SubTensor` anymore in this example.
```
ezyang: Under my proposal, `__torch_function__` never dispatches to `SubTensor` methods ever; instead, `__torch_function__` is the base case that gets us to call `func._implementation` (terminates `__torch_function__` handling)

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,358 @@
...
+In this instance, with the proposed changes, processing would follow the
+`__torch_function__` protocol. This means that control would end up in
+`SubTensor.__torch_function__`, go to `Tensor.__torch_function__` from there
+and and then come to `SubSubTensor.__add__`, from where it would go to
+`Tensor.__add__`, and then back up the stack in the reverse order. This means
+that great care needs to be taken when writing `SubTensor.__torch_function__`
+to take into account the fact that it has to handle subclass methods.
+
+### Protocol support for external libraries
+We will also recommend that all `Tensor` subclasses make their own methods go
```
ezyang: It's worth clarifying that this describes *completely new methods* which aren't present on `Tensor`; the reasoning here is equivalent to how we can annotate Python level functions with the dispatch decorator and trigger `__torch_function__` handling (we got rid of the decorator to speed up core code, but for end users this is perfectly fine.)

ezyang:
One more detail which is not currently explained in the RFC:  How are we going to manufacture `func` for methods passed to `__torch_function__`? In the current `__torch_function__` implementation these correspond exactly to the actual functions that the user calls from top level, e.g., `torch.add`. So, I suppose these should be *unbound* methods like `torch.Tensor.__add__`? In that case, `self` should be present in `args`? But now there is a hazard; the user must be sure to use the `self` from `args`, and not the top level self, if this is a multi-dispatch Tensor, because it's possible the user did something like `plain_tensor.add(my_subclass_tensor)` and thus the `self` of `__torch_function__` corresponds to the non-self argument, not self (which is `plain_tensor`). This is harder to get wrong with the old `__torch_function__` because there are only functions involved. (Maybe this should have been a `staticmethod`?)

RFC-0001-torch-function-for-methods.md
hameerabbasi: The main reason for having the `types` in here was threefold:

* Do not mix unrelated class trees. Removing this would.
* Provide an easy way for a `__torch_function__` implementation to check if it can support the supplied types.
* Signal that dispatch has to go through a different path (now defunct, so I agree with you).

Expanding on each of these a little,

### Do not mix unrelated class trees
Suppose there is `Tensor`, `SubTensorA`, and `SubSubTensorA` that form an inheritance chain, and `Tensor`, `SubTensorB` and `SubSubTensorB` that form another inheritance chain. If we remove this class, `SubTensorA + SubSubTensorA` would return `SubSubTensorA` (good, subclasses come before superclasses), but `SubTensorB + SubTensorA` would return `SubTensorB` (oookay...) and `SubTensorA + SubTensorB` would return `SubTensorA` (a-ha! inconsistency found!)

With this part of the proposal in place, both of the above would raise `TypeError`.

This applies not just to `Tensor` subclasses but also all kinds of mixing. See below.

### Check for support
Right now, the `Tensor`-likes are buried inside the `args` and `kwargs`. Now, envision a future where we have implementations of `__torch_function__` on a lot of classes, possibly being mixed.

We faced exactly this dilemma in `__array_function__`, and `types` was the solution.

Assume `DuckTensorA` and `DuckTensorB` are two such classes. Now envision a fairly complex function like `torch.cat`, or the equivalent of [`numpy.block`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.block.html#numpy.block).

How are `DuckTensorA` and `DuckTensorB` to know that they don't support each other, without going recursively into each argument and extracting the type (which is fatal for performance, btw)? Why, they just inspect `types`!

hameerabbasi:
> Maybe this should have been a `staticmethod`?

We need a way to copy metadata from `self`, which isn't on `type`.

RFC-0001-torch-function-for-methods.md
hameerabbasi: It does.

ezyang:
> We need a way to copy metadata from self, which isn't on type.

Sure, but if I do a binary operation between two SubTensors, I in general need to consider the metadata on obth of them. Self isn't enough!

RFC-0001-torch-function-for-methods.md
ezyang: > Do not mix unrelated class trees

I agree that we should not mix unrelated class trees. I wonder if the "reference" implementation on Tensor (that rewraps subclasses) can directly determine if class tree mixing would occur by looking at the arguments. It already needs to extract types from the tensor like arguments because we need to determine which `__torch_function__` to invoke for multiple dispatch; so it can also easily check if all the subclasses are related or not. Is there a use case for user-defined `__torch_function__` to also do this calculation?

> Check for support

Yeah, no counterargument here. :) Alright, types can stay :> (however, I'm still not convinced you have to strip out types below you in the subclass hierarchies on super calls!)

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,413 @@
...
+`__torch_function__` protocol. This means that control would end up in
+`SubTensor.__torch_function__`, go to `Tensor.__torch_function__` from there
+and and then come to `SubSubTensor.__add__`, from where it would go to
+`Tensor.__add__`, and then back up the stack in the reverse order. This means
+that great care needs to be taken when writing `SubTensor.__torch_function__`
+to take into account the fact that it has to handle subclass methods.
+
+In general, control flow will follow this pattern:
+
+![Control flow diagram](./RFC-0001-assets/dispatch-flow.svg)
```
rgommers: Maybe the second hand-drawn diagram from @ezyang  should be added here as well as an svg. The text above about control flow for `__add__` is hard to follow without it.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,413 @@
...
+
+class SubSubTensor(SubTensor):
+    def __add__(self, other):
+        # Pre-processing
+        ret = super().__add__(other)
+        # Post-processing
+        return ret
+```
+
+In this instance, with the proposed changes, processing would follow the
```
rgommers: try not to use "would ..." or "with proposed changes" - it's not clear if this is what _will_ happen, or if you're discussing some discarded alternative. just phrase it like "it works like a, b, c"

RFC-0001-torch-function-for-methods.md
hameerabbasi: That one's rendered invalid since essentially decorating an overridden method isn't something we do anymore.

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,413 @@
...
+
+```python
+class SubTensor(torch.Tensor):
+    @classmethod
+    def __torch_function__(cls, func, types, args, kwargs):
+        # Implementation here
+```
+
+The reason for adding `types` to the signature is necessitated by the need for
+`super()`. If we set a requirement for `super().__torch_function__` to work
```
ezyang: This is not relevant anymore, right?

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,413 @@
...
+In addition, there has been interest in adding a "universal hook" that operated
+on both functions and methods, perhaps modifying the control flow before
+returning the result. Such a hook already exists today in the form of
+`__torch_function__`, however, it only operates on functions and not on
+methods, and support for subclassed `torch.Tensor` objects in this protocol is
+limited.
+
+### Proposal
+We propose the following signature change to `__torch_function__`, to make it
+match NumPy: [[4]]
```
ezyang: The `classmethod` here doesn't match numpy, right? :)

RFC-0001-torch-function-for-methods.md
```diff
@@ -0,0 +1,413 @@
...
+        return ret
+```
+
+This method has the effect of passing through subclasses through all
+functions/methods as intended.
+
+This corresponds exactly to the implementation `numpy.ndarray` gains in [[4]],
+except for the fact that subclasses are passed through via another internal
+mechanism (namely the `__array_finalize__` protocol) there, as well as the fact
+that we are checking subclassing against `type(self)` instead of `Tensor`. This
```
ezyang: out of date now

Balandat:
This is very interesting, thanks for the RFC. Am I right to assume this could go a long way in helping enable something like a `LinearOperator` abstraction that is described in https://github.com/pytorch/pytorch/issues/28341 ?

hameerabbasi:
@Balandat I cannot comment more broadly, but certainly, the code example you show in the issue description should be possible, yes. With my understanding of the problem it should all be possible.

RFC-0001-torch-function-for-methods.md
rgommers: @hameerabbasi capturing the above comments about "teaching moment" in a code comment would be useful.

> This kind of suggests we need either (1) a way to name functions that bypasses dispatch, or (2) a way to invoke a function without triggering dispatch. I'm guessing this is what the `_implementation` field on these functions does; worth calling this out explicitly in the RFC!

So you did this by providing a (non-public?) context manager, right @hameerabbasi?

hameerabbasi: That's correct, that's the conclusion we came to in a call with @ezyang 

The upshot of it was that we're not overly concerned about thread forks, the state of `__torch_function__` can be captured within a thread-local variable, and there aren't any forks between the moment the function is called and dispatching, where `__torch_function__` is checked.

This is also the reason it's private -- the context doesn't survive when new threads are spawned. pytorch/pytorch#28520 may resolve this.

rgommers:
@hameerabbasi now that the code landed, can you make final updates to this RFC to adjust to reality? Then let's merge it.

RFC-0001-torch-function-for-methods.md
```diff
@@ -407,7 +407,114 @@ To implement this proposal requires three main steps:
...
+implementations of particular functions.
+
+### Functions vs Methods vs Properties
+Both functions and methods/properties on `torch.Tensor` will be possible arguments to
+`__torch_function__`. These are different in subtle but important ways, and
+in some cases it is required to handle them differently. For instance,
+`torch.Tensor` methods/properties have the following properties:
+
+1. They can only accept `torch.Tensor` instances as the first argument.
+2. They *may or may not* have a `__module__` defined.
```
ezyang, line 469: This kind of comes out of the blue; maybe this would be better in a FAQy or appendix section

rgommers (approved):

rgommers:
Okay, let's finally get this in! Thanks @hameerabbasi and @ezyang

turian:
Has this RFC been implemented?

Related: https://github.com/pytorch/pytorch/issues/52265

ezyang:
Yup!

PR #4, opened 2020-04-08
Proposed Roadmap for torch.sparse matrix multiplication 

dhavide:
Suggestions for a strategy on how to go forward with `torch.sparse`

- This doc discusses only matrix multiplication (because it is mentioned in many issues)
- The goal is to start here as a way to go deeper into other issues associated with `torch.sparse`.

### DIFF
--- a/PyTorch-Sparse-Roadmap.md
+++ b/PyTorch-Sparse-Roadmap.md
@@ -1,0 +1,440 @@
+# Roadmap for `torch.sparse` Matrix Product API
+
+|            |                 |
+| ---------- | --------------- |
+| Authors    | Dhavide Aruliah |
+| Status     | Draft           |
+| Type       | Process         |
+| Created    | 2020-04-07      |
+| Resolution | TBD             |
+
+
+## Abstract
+
+At the time of writing, the 
+[`torch.sparse`](https://pytorch.org/docs/stable/sparse.html) module API is 
+considered experimental. There are more than 60 unresolved GitHub 
+issues involving `torch.sparse` to address&mdash;some dating back to 
+2018&mdash;and a handful of which have been flagged as high priority. About
+twenty of those issues deals with *multiplication* of sparse PyTorch
+matrices and at least one of those is a high priority issue.
+
+This proposal summarizes an overarching strategy for addressing a number 
+`torch.sparse` issues related to the *performance of sparse matrix multiplication* in 
+PyTorch. There are, of course, many other `torch.sparse` issues flagged on 
+GitHub that point to other desired features & API design considerations; these 
+particular matrix multiplication issues are a useful starting point for further 
+API improvements for this module. That is, dealing with these issues first can
+seed development of additional new `torch.sparse` features.
+
+## Motivation and Scope
+
+Almost one third of the current `torch.sparse` issues relate to the timings of 
+matrix multiplication with sparse matrices (i.e., sparse tensors of rank 2) in 
+PyTorch (e.g., issue [\#5262](https://github.com/pytorch/pytorch/issues/5262)). 
+There are additional issues relating to batching sparse matrix products (e.g., 
+issue [\#33430](https://github.com/pytorch/pytorch/issues/33430)) and computing 
+gradients (e.g., issue 
+[\#12498](https://github.com/pytorch/pytorch/issues/12498)) but these can 
+likely be addressed *after* the performance issues. Note also that some issues 
+that refer to "multiplication" explicitly are unrelated in that they refer to 
+*elementwise* products & *broadcasting* with sparse tensors (e.g., issue 
+[\#2909](https://github.com/pytorch/pytorch/issues/2909)).
+
+For context, the internal representation of PyTorch sparse tensors uses
+[*COO* (*coordinate*) format](https://en.wikipedia.org/wiki/Sparse_matrix#Coordinate_list_(COO)),
+i.e., using a rank 1 `FloatTensor` of length `nnz` for the non-zero values and 
+a `nnz x d` `IntTensor` to store the positions of those entries in a sparse 
+tensor of rank `d`:
+
+```{python}
+>>> i = torch.LongTensor([[0, 1, 1],
+                          [2, 0, 2]])
+>>> v = torch.FloatTensor([3, 4, 5])
+>>> torch.sparse.FloatTensor(i, v, torch.Size([2,3])).to_dense()
+ 0  0  3
+ 4  0  5
+[torch.FloatTensor of size 2x3]
+```
+
+One unusual feature of the PyTorch COO representation of tensors is that it 
+permits a *hybrid* of sparse and dense dimensions. That is, the leading
+dimensions can be sparse and the remaining dimensions are dense; the methods
+`sparse_dim` and `dense_dim` reveal those dimensions explicitly. Another way
+to think of this is as a (sparse) tensor with (dense) tensors as entries.
+
+It is important to observe that the `torch.sparse` API currently supports 
+*uncoalesced* sparse tensors (i.e., sparse tensors in COO format with duplicate 
+coordinates in the indices). Values associated with repeated indices in an
+uncoalesced sparse tensor are added together to yield the value associated with
+that index. Permitting uncoalesced sparse tensors improves the 
+efficiency of certain operations (notably addition of sparse tensors and 
+insertion of new nonzero elements into a sparse tensor).
+
+For instance, in computing `(A + B + C) * D` for some tensors `A`, `B`, `C`,
+& `D` of identical shape, the uncoalesced sum `(A + B + C)` can be computed
+by concatenating the lists of indices & values from all three tensors. To
+compute the elementwise product with `(A + B + C) * D`, the tensor
+`(A + B + C)` must be coalesced first; this involves sorting the indices and
+summing values associated with repeated indices). By contrast, forcing
+sparse tensor operations to yield coalesced sparse COO tensors requires
+unnecessary additional work: sorting indices to compute `(A + B)` and then
+again to compute `(A + B) + C`.
+
+A number of sparse linear algebra libraries&mdash;including
+[`scipy.sparse`](https://docs.scipy.org/doc/scipy/reference/sparse.html)&mdash;permit
+the use of
+[*CSR* (*compressed sparse row*)](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format))
+format to represent sparse matrices as well. The CSR sparse format enables
+certain known efficient algorithms for BLAS operations of Level 2 & 3.
+Indeed, the SciPy documentation for the
+[`csr_matrix`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html#scipy.sparse.csr_matrix)
+class implies that COO format is mainly useful for assembling sparse matrices
+(e.g., in the context of finite element analysis) while CSR format is more
+useful for sparse *matvecs* (*matrix-vector products*, Level 2 BLAS operations)
+and sparse matrix-matrix products (Level 3 BLAS operations).
+
+Lower-level libraries like
+[Intel's MKL](https://software.intel.com/en-us/mkl-developer-reference-fortran-blas-and-sparse-blas-routines)
+provide sparse BLAS algorithms using numerous internal representations including
+COO & CSR. On CUDA architecture GPUs,
+[cuSPARSE](https://developer.nvidia.com/cusparse),
+[CUSP](https://developer.nvidia.com/cusp), and
+[MAGMA](https://ceed.exascaleproject.org/magma/#magma-sparse)
+all support sparse matrix computations in both CSR & COO formats.
+There are
+[benchmarks](https://pdfs.semanticscholar.org/5478/de620de99b1d4e41599a8950c7d3d9ff07b5.pdf)
+that suggest that sparse matvecs can be between 2 and 2.5 times faster when
+using CSR format over COO format. Of course, with uncoalesced COO format,
+there are additional concerns about deterministic computations on GPUs that
+arise due to nonuniqueness of the internal representation in both the sequence
+and contents of the index and values arrays.
+
+Numerous GitHub issues&mdash;e.g., issue
+[\#16187](https://github.com/pytorch/pytorch/issues/16187)
+that presents timed comparisons of PyTorch & SciPy&mdash;draw attention to
+performance deficiencies of PyTorch sparse matrix products. The precise causes
+are not obvious & require a fairly deep exploration of the internal PyTorch
+implementation. The performance penalty could be caused by conversions from
+COO to CSR format, from coalescing sparse COO tensors, from
+not using optimized libraries that employ, e.g.,
+[sparse BLAS](https://math.nist.gov/spblas/), or some combination of these.
+[This Stack Overflow
+question](https://stackoverflow.com/questions/30118673/why-cusparse-is-much-slower-than-cublas-for-sparse-matrix-multiplication)
+provides some additional discussion of why the cause of this performance
+penalty is difficult to assess.
+
+With the considerations above, it makes sense to explore extending
+`torch.sparse` to provide support storing sparse matrices in CSR format
+(in the special case of tensors of rank 2 or with `sparse_dim()==2`). There
+are two alternative schemes to consider:
+
++ providing user-accessible conversion routines between COO & CSR 
+representations for rank 2 tensors (or tensors with `sparse_dim()==2`); and
++ using a cached internal CSR representation.
+
+The first approach is most obvious; its primary disadvantage is that 
+supporting alternative storage formats requires modifying large portions of the 
+code obscured by the API (because, e.g., the algorithms for elementwise
+arithmetic operations differ for CSR & COO formats). These code overheads can
+be avoided using the second option instead, i.e., by caching. However, caching
+a CSR representation of a sparse tensor on top of the COO representation
+effectively doubles the storage required and introduces additional complications
+in having to update the cached representation on modification of any individual
+element. Assuming that the caching problems are worse, we'll consider the first
+approach only in this proposal.
+
+The main goals of this proposal, then, are:
+
+1. to implement a CSR representation for `torch.sparse` tensors of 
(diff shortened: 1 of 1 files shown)


ezyang:
One very basic newbie question: is there a generalization of CSR format that works for higher dimensions? Or is it a 2D only kind of deal?

gchanan:
I did a quick read, thanks for the thorough writeup!

Some questions that come to mind:
1) Can the csr format be generalized to more than just 2-d tensors?  Having "2d" in all the names feels pretty clunky and we could just error out if the input matrix is not 2d if it's not generalizable.
2) I'd like to see more detail on the constructor(s). The conversion operator IMO doesn't really matter because it's mostly used in tests (most uses for sparse tensors you wouldn't want to materialize the dense tensor anyway).

hameerabbasi:
> One very basic newbie question: is there a generalization of CSR format that works for higher dimensions? Or is it a 2D only kind of deal?

There is GCRS (https://ieeexplore.ieee.org/document/7237032) and CSF (https://dl.acm.org/doi/10.1145/2833179.2833183). If you want to go completely-bonkers-general-all-formats-are-one there's TACO (https://arxiv.org/abs/1804.10112).

albanD:
Hi,

Thanks for the proposal.
I would have one question about the timeline.
I assume that this new Tensor type will have `.is_sparse() == true`. But how is it going to interact with code that already exists and use this to special case on the COO sparse Tensors we have?
For example here https://github.com/pytorch/pytorch/blob/5bbcddae3bfb7abaeaada29369c6ed5964d390fe/torch/csrc/autograd/functions/accumulate_grad.h#L61-L68 we assume that when we have a sparse Tensor, then we can access _values and _indices. But this might not be true anymore.
Should a new point be added to the timeline to investigate the current codebase and make sure all the special casing we do for sparse Tensors will be ok for all sparse Tensors?

dhavide:
> I did a quick read, thanks for the thorough writeup!
> 
> Some questions that come to mind:
> 
>     1. Can the csr format be generalized to more than just 2-d tensors?  Having "2d" in all the names feels pretty clunky and we could just error out if the input matrix is not 2d if it's not generalizable.
> 
>     2. I'd like to see more detail on the constructor(s). The conversion operator IMO doesn't really matter because it's mostly used in tests (most uses for sparse tensors you wouldn't want to materialize the dense tensor anyway).

1. As Hameer points out, there are generalizations beyond 2D. We actually do need to consider this
    because there are use cases (e.g., NLP) in which storing batches of sparse matrices is useful.
    The precise implementation of this is quite complicated and needs to be considered carefully.
    We can modify the proposal to make this more explicit.

    Anyway, it's fine with to change the identifiers (someone else has already winced at the `2d`
    suffix for the modifications to the API). The primary motivation was to serve as a warning of a
    restriction, but that restriction may not exist in the eventual implementation.
2. More details about the constructor is a good idea. We'll modify the proposal accordingly.

dhavide:
> Hi,
> 
> Thanks for the proposal.
> I would have one question about the timeline.
> I assume that this new Tensor type will have `.is_sparse() == true`. But how is it going to interact with code that already exists and use this to special case on the COO sparse Tensors we have?
> For example here https://github.com/pytorch/pytorch/blob/5bbcddae3bfb7abaeaada29369c6ed5964d390fe/torch/csrc/autograd/functions/accumulate_grad.h#L61-L68 we assume that when we have a sparse Tensor, then we can access _values and _indices. But this might not be true anymore.
> Should a new point be added to the timeline to investigate the current codebase and make sure all the special casing we do for sparse Tensors will be ok for all sparse Tensors?

Excellent point. Yes, this has the potential to break existing parts of the codebase, so an additional point needs to be added to the timeline to verify this (concerns about `autograd` are mentioned in passing in the current draft, but this certainly needs to be called out explicitly in the timeline).

PyTorch-Sparse-Roadmap.md
```diff
@@ -0,0 +1,440 @@
...
+2. Construct and test a prototype implementation of conversion functions
+   `torch.to_sparse_csr2d` and `torch.to_sparse_coo` to convert sparse PyTorch
+   tensors of sparse rank 2 between COO and CSR formats.
+
+   + The implementation of related functions in 
+     [Tensorflow](https://www.tensorflow.org/api_docs/python/tf/raw_ops/SparseMatrixSparseMatMul)
+     can serve as a useful model for these transformations.
+   + Include exception handling to ensure `sparse_dim()==2` before converting to from COO to CSR.
+   + See also comments in issue
+     [\#16187](https://github.com/pytorch/pytorch/issues/16187).
```
ezyang, line 272: NB: conversion functions need to work on CUDA :>

PyTorch-Sparse-Roadmap.md
```diff
@@ -0,0 +1,440 @@
...
+   + See [issue # 16187](https://github.com/pytorch/pytorch/issues/16187)
+     for representative examples of benchmarks.
+   + Include examples of matrices of fairly large dimensions (i.e., where dense
+     representations are impractical or infeasible).
+   + Include examples with `sparse x dense` and `sparse x sparse` matrices for
+     the timings.
+
+2. Construct and test a prototype implementation of conversion functions
+   `torch.to_sparse_csr2d` and `torch.to_sparse_coo` to convert sparse PyTorch
+   tensors of sparse rank 2 between COO and CSR formats.
```
ezyang, line 265: If CSR is going to be another layout type, there is some infrastructural work that is going to need to be done here; you need a new tensor type (since a CSR isn't going to have the same memory representation as COO tensor). For example, right now we refer to COO sparse as just "Sparse" in the codebase; it might make sense to disambiguate these names to "SparseCOO" so people don't mix them up. (Or, I suppose, we can just grandfather sparse to mean COO sparse, and be explicit in all the new locations).

pearu: The identity "Sparce == Sparce COO" is not consistent in current pytorch, for instance, there are sparse tools that include `coo` explicitly in their names: `torch.sparse_coo_tensor`, `at::_sparse_coo_tensor_with_dims`, `torch.sparse_coo`, etc.
This indicates that the design of the current infrastructure implicitly would support other sparse formats. For instance, one could have  `torch.sparse_csr_tensor`, etc.

rgommers:
> > One very basic newbie question: is there a generalization of CSR format that works for higher dimensions? Or is it a 2D only kind of deal?
> 
> There is GCRS (https://ieeexplore.ieee.org/document/7237032) and CSF (https://dl.acm.org/doi/10.1145/2833179.2833183). If you want to go completely-bonkers-general-all-formats-are-one there's TACO (https://arxiv.org/abs/1804.10112).

Important disclaimer there is that those >2-D generalizations are not supported by cuSPARSE, MKL et al. CSR is the best-supported format by those libraries, followed by COO. So it's unclear to me whether it makes sense to have more than one >2-D format.

PyTorch-Sparse-Roadmap.md
dhavide: Noted. Appropriate modification made in revised proposal.

PyTorch-Sparse-Roadmap.md
dhavide: I think Pearu's comment is correct; it is possible to extend the current sparse classes without having to specify the storage format by a name like "SparseCOO." This could perhaps be achieved by adding a `_layout` attribute of type `torch.sparse_coo` or `torch.sparse_csr` or some other valid PyTorch object of class `torch.layout`.

The proposal has been modified to provide a more thorough description of the CSR format and the associated constructor.

PyTorch-Sparse-Roadmap.md
```diff
@@ -0,0 +1,515 @@
...
+
+   + See [issue # 16187](https://github.com/pytorch/pytorch/issues/16187)
+     for representative examples of benchmarks.
+   + Include examples of matrices of fairly large dimensions (i.e., where dense
+     representations are impractical or infeasible).
+   + Include examples with `sparse x dense` and `sparse x sparse` matrices for
+     the timings.
+
+2. Construct and test a prototype implementation of conversion functions
+   `torch.to_sparse_csr2d` and `torch.to_sparse_coo` to convert sparse PyTorch
```
pearu, line 323: Change: `to_sparse_csr2` -> `to_sparse_csr`

PyTorch-Sparse-Roadmap.md
```diff
@@ -0,0 +1,515 @@
...
+  conversion between representations (for rank 2 sparse tensors).
+
+The `torch.sparse_csr_tensor` constructor would have a similar signature to `torch.sparse_coo_tensor`, i.e.,
+
+```{python}
+sparse_csr_tensor(indices, values, size=None, dtype=None,
+                  device=None, requires_grad=False)       -> Tensor
+```
+
+where `indices` would be a list of two lists (e.g., `row_index` & `col_index`
```
pearu, line 306: Does the construction of CSR from `indices, values` allow N-D generalization? E.g. when using GCRS or CSF formats?
If yes, then +1. If not, what would be the uniting model for constructing CSR and GCRS?

PyTorch-Sparse-Roadmap.md
```diff
@@ -0,0 +1,515 @@
...
+from the example before) and `values` would be a tensor containing the
+corresponding nonzero entries of the sparse tensor.
```
pearu, line 308: Remove "nonzero" or replace it with "presumably nonzero" as nothing would check if the entries will be nonzero (it just would be too expensive to do that).

Should we be explicit here on whether the dense tensor of `values` can be an N-dimensional tensor o(similar to the COO case) or not?

PyTorch-Sparse-Roadmap.md
pearu: Or would it make sense to introduce a new layout for each generalization instead?

ezyang:
I think we've already gotten here in the convo, but just to state it explicitly: if csr unambiguously refers to 2d only, we don't need to qualify the name with 2d; and we don't have to work on generalized versions if it's not clear, 2d only is fine.

albanD:
Wouldn't a batch of sparse matrix be an important usecase as well? 
For nn use, have only 2D Tensors will be quite limited.

dzhulgakov:
This is great proposal! Interesting thought: CSR is effectively equivalent to `offsets` representation that embedding_bag and other operations use. So it'd be good to explore whether we can change APIs there too to accept sparse tensors.

I'd also vote for extending representation to >2 dimensions if possible (or at least not baking it in API names now). Batch use case Alban mentioned is one example. Another point is that several embedding_bags' input can be thought as sparse 3d tensor (e.g. batch X num_features X feature_dim for sparse features). So it might be interesting to explore whether sparse tensor api is a good representation for it.

PyTorch-Sparse-Roadmap.md
dhavide: Whoops, fixed now. Thanks.

PyTorch-Sparse-Roadmap.md
dhavide: Okay, I've modified the description of CSR to N dimensions to include a specification of how to do so. Basically, for a sparse tensor of rank `d>=2` whose leading dimension has length `M`, the CSR format requires storing three arrays: `RO` (an integer array of length `M+1`), `CO` (an integer matrix of dimension `(d-1) x nnz`), and `VL` (an array of values of length `nnz`).

I think that this generalization is sufficient (even if the indexing sparse dimension is always the leading dimension). Permitting a new layout for other specific generalizations is available for those who choose to do so.

Representing batches of sparse matrices is required in numerous contexts (e.g., a large number of embedding matrices in NLP). The obvious Python way to do this would be in lists or dicts or standard containers, but, depending on the size of the batches, performance issues are an obvious problem.

PyTorch-Sparse-Roadmap.md
dhavide: Fixed.

dhavide:
> Wouldn't a batch of sparse matrix be an important usecase as well?
> For nn use, have only 2D Tensors will be quite limited.

@albanD  yes, batches of sparse matrices is obviously important. I have modified the proposal to describe how CSR extends to tensors of arbitrary rank (and added references to papers that describe other extensions in greater detail). I believe this category of generalization is sufficient (even if the indexing sparse dimension in `RO` is always the leading dimension).

pearu:
@dhavide could you also rename `PyTorch-Sparse-Roadmap.md` to `PyTorch-Sparse-MatMul-Roadmap.md`, or similar?

jamestwebber:
@rgommers pointed me here from the CuPy issue–glad to see progress being made! To add some comments from an end-user (domain is computational biology).

> The big assumption is that faster sparse matrix products actually constitutes a meaningful benefit to PyTorch users—the hope is that the accuracy of this assumption will become clear in the early stages of prototyping.

I think this would have a major benefit in my area of research. I work a lot with single-cell RNA sequencing data (scRNAseq), and the raw data is a large, sparse count matrix. By "large": on the order of 10k rows for one experiment, easily 100k to 1M+ for very large projects or meta-analyses, and 10s of thousands of features. By "sparse": typically 1-2% nonzero values. Not "big data" so much as wide, complex data. This type of data is becoming more prevalent and datasets are growing as methods improve, see [this google sheet](https://docs.google.com/spreadsheets/d/1En7-UV0k0laDiIfjFkdn7dggyR7jIk3WH8QgXaMOZF0/edit#gid=1444925081) showing the trend over time (credit to @vals who might drop in later).

When analyzing scRNA data, the difference between sparse and dense representations can be the difference between fitting the data in GPU memory or not, which can greatly impact the speed of training. The rest of the model could be dense but just using sparse tensors for the input would allow for larger batches and less memory-copying. It's hard to know exactly how big a benefit that would provide but I think it'd be considerable.


One thing I would add as something to think about (but out of scope here, I guess): I'd love to be able to pass a sparse tensor in to `torch.distributions`. For example, call `torch.distributions.Binomial(total_counts=sparse_counts, p=0.5).sample()` and get a sparse tensor with the same shape and sampled counts in the nonzero locations.

That's kind of a niche thing to do but it might be doable with the `fill_value` attribute, for distributions where certain values have a trivial answer. My hunch is that most users want sparse tensors to be a drop-in replacement as much as possible, because trying to keep track of the options gets difficult (I always have this issue with scipy, where the sparse datatype is always a matrix, not an array).

RFC-0004-pyTorch-sparse-matmul-roadmap.md
```diff
@@ -0,0 +1,544 @@
...
+The `torch.sparse_csr_tensor` constructor would have a similar signature to `torch.sparse_coo_tensor`, i.e.,
+
+```{python}
+sparse_csr_tensor(indices, values, size=None, dtype=None,
+                  device=None, requires_grad=False)       -> Tensor
+```
+
+where `indices` would be a list of two lists (e.g., `RO` & `CO`
+from the description of CSR format above) and `values` would be a tensor
+containing the corresponding (presumably) nonzero entries of the sparse
```
pearu, line 332: This API for CSR is incorrect/inefficient: `RO` has a length `size[0]+1` while `CO` has length `nnz` and so `RO` and `CO` lists should be separated.

I suggest the following API for `sparse_csr_tensor`:
```python
sparse_csr_tensor(pointers, indices, values, size=None, dtype=None,
                  device=None, requires_grad=False)       -> Tensor
```
where the `pointers` argument is a 1-D array of length size[0] + 1 containing the cumulative number of nonzero elements in the sparse array computed for each index of the first sparse dimension.

Xref: https://github.com/Quansight/pearu-issues/issues/28

RFC-0004-pyTorch-sparse-matmul-roadmap.md
```diff
@@ -0,0 +1,544 @@
...
+   + See [issue # 16187](https://github.com/pytorch/pytorch/issues/16187)
+     for representative examples of benchmarks.
+   + Include examples of matrices of fairly large dimensions (i.e., where dense
+     representations are impractical or infeasible).
+   + Include examples with `sparse x dense` and `sparse x sparse` matrices for
+     the timings.
+
+2. Construct and test a prototype implementation of conversion functions
+   `torch.to_sparse_csr` and `torch.to_sparse_coo` to convert sparse PyTorch
+   tensors of sparse rank 2 between COO and CSR formats.
```
pearu, line 351: I think we can target immediately arbitrary rank (>=2) CSR sparse tensors in the constructor, including the hybrid feature, without any extra effort. The algorithms, however, may be restricted to 2-D in the short run.

RFC-0004-pyTorch-sparse-matmul-roadmap.md
```diff
@@ -0,0 +1,544 @@
...
+respectively contain the extents of rows, the column indices, and the numerical
+values. That is, to represent an `M x N` sparse matrix `A` with `nnz` nonzero
+entries (floating-point) in CSR format requires three separate rank 1 arrays:
+`RO` (an integer array of length `M+1`), `CO` (an integer array of length
+`nnz`), and `VL` (a floating-point array of length `nnz`). In the
+extension of CSR format to tensors of rank greater `d>2`, the array `CO`
+would be a matrix of dimensions `(d-1) x nnz`.
+
+The entry in `RO[r]` is the cumulative number of nonzero entries in the matrix
+up to and including row `r` (`r=0...M+1`). Thus, the first entry of `RO` is `0`
```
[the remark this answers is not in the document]
pearu: Fixed.

RFC-0004-pyTorch-sparse-matmul-roadmap.md
```diff
@@ -0,0 +1,544 @@
...
+and the last&mdash;`RO[M+1]`&mdash;is `nnz`. The arrays `CO` and `VL` of length
```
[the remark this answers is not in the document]
pearu: Fixed.

npuichigo:
So will block sparse be supported as some people ask for that maybe for sparse attention? (https://github.com/pytorch/pytorch/issues/9222)

rgommers:
Thanks for asking @npuichigo. We've discussed it briefly recently and it's clearly of interest. The first target for increasing sparse matrix performance at the moment is the GCS format: https://github.com/pytorch/pytorch/pull/44190. Considering the block sparse use cases and interest in https://github.com/pytorch/pytorch/issues/9222, it does seem like that's a good next candidate. I'll add a note to the tracking issue for sparse tensor improvements (https://github.com/pytorch/pytorch/issues/44634) to reassess the priority of block-sparse.

Issue #5, opened 2020-04-27
Roadmap for PyTorch Sparse Tensors


pearu:
This issue is created to reserve a RFC number for the subject.

rgommers:
We don't need this issue, so I'll close it and put the RFC number in the PR title.

PR #6, opened 2020-04-27
RFC0005 - Roadmap for PyTorch Sparse Tensors [WIP]

pearu:
Tackles #5 .

### DIFF
--- a/RFC0005-sparse-roadmap/README.md
+++ b/RFC0005-sparse-roadmap/README.md
@@ -1,0 +1,139 @@
+# Roadmap for PyTorch Sparse Tensors
+
+|            |                 |
+| ---------- | --------------- |
+| Authors    | Pearu Peterson  |
+| Status     | Draft           |
+| Type       | Process         |
+| Created    | 2020-04-23      |
+| Resolution | TBD             |
+
+
+## Abstract
+
+The aim of this document is to define a direction of PyTorch sparse
+tensor features: enable a sparse tensor to be a drop-in replacement of
+dense tensors.
+
+
+## Motivation and Scope
+
+PyTorch implements the Tensor object as the basic structure of storing
+data in a multi-dimensional array structure that can be processed on
+various computational devices such as GPUs as well as multicore CPUs
+efficiently by taking advantage of parallelism that these devices
+support. PyTorch has advanged in developing computational tools for
+various ML and AI algorithms that can be applied to so-called dense or
+strided tensors. However, the developement of tools applicable to
+sparse tensors has not passed the experimental stage after two or more
+years of development, dispite the fact that a sparse tensor would be a
+natural data storage format for many ML&AI applications where the data
+characterizes some pairwise relations between study subjects or when
+the aim is to analyze only a small set of features that dominate in a
+big system, for instance. Choosing a sparse tensor format may lead to
+more performant programs due to the efficient uses of memory as well
+as computational resources.
+
+The starting point of this proposal and the far-reaching goal is that
+the PyTorch tensor represents a multi-dimensional array whereas the
+used data storage format is an implementation detail that users can
+select depending on their data to (indirectly) control the choise of
+computational tools that would be most efficient for users'
+applications. Switching a storage format of data to another data
+format (dense to sparse, sparse to dense, sparse to sparse, etc)
+should not require changes to the data processing algorithms written
+by the end-users or other high-level software tools. This approach of
+supporting data storage format agnostic algorithms allows faster
+adoption of newly developed algorithms and mathematical models as
+software programs.
+
+PyTorch already implements a dispatcher system for executing programs
+on a particular device depending on the data storage location. The
+same dispatcher system can be used for selecting computational tools
+depending on the data storage format. So, the implementation of this
+proposal becomes about (i) providing sparse tensor implementations and
+the corresponding low-level computational tools, and (ii) reusing
+PyTorch dispatcher system to make sparse tensors as drop-in
+replacements to existing dense tensors when used as inputs to
+high-level computational tools. A prioritization of features sets is
+required in order to make the stable sparse tensors support accessible
+for users' programs as soon as possible.
+
+
+## Mathematical background of arrays
+
+One of the most common data representation in scientific and
+engineering computations is an array of items that share the same data
+type and size and that can be generalized to any dimensions. The array
+items are naturally indexed using tuples of integers that represent
+the coordinates in a multidimensional discrete space of data values.
+Mathematically, arrays can be interpreted as vectors, matrices, or
+tensors that form algebras with various algebra operations as well as
+functions on algebra elements.
+
+When representing data as an array in computer memory, a array storage
+format must be selected. The simplest and most general array storage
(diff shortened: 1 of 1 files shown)


ezyang:
Some high level feedback: this proposal is pretty long, and I know there are parts that are not directly relevant to sparse tensors; for example a, multipage section on traditional dense tensors (https://github.com/pytorch/rfcs/tree/9a9c57c21b6e7ee464661572a50b6e4c4c608d8f/RFC0003-sparse-roadmap#strided-array-storage-format). This makes it harder for external readers (like me) to see what the substantive parts of the proposal are. For example, way in https://github.com/pytorch/rfcs/tree/9a9c57c21b6e7ee464661572a50b6e4c4c608d8f/RFC0003-sparse-roadmap#large_blue_circlefill-value-api we see that you propose also supporting undefined fill. That's a pretty major proposal (partially because it contradicts "a sparse tensor is a drop-in replacement of dense tensors for all tensor operations"--undefined fill sparse tensors are not convertible with dense tensors, in general it's not possible to losslessly go from an undefined fill sparse tensor to dense tensor and then back again).

rgommers:
Thanks @ezyang. I brought that up recently as well; this proposal has been growing and growing. We decided it'd be better to take individual parts of it and resubmit them as focused individual proposals or PRs. The first ones should be:
- A PR to the PyTorch docs with the tables of which functions support sparse tensors
- A new RFC for only the fill value feature

pearu:
> ... we see that you propose also supporting undefined fill. That's a pretty major proposal (partially because it contradicts "a sparse tensor is a drop-in replacement of dense tensors for all tensor operations"--undefined fill sparse tensors are not convertible with dense tensors, in general it's not possible to losslessly go from an undefined fill sparse tensor to dense tensor and then back again).

Thank you, @ezyang , for this observation. There is a one-to-one correspondence between dense tensors and sparse tensors with defined fill-value. Extending the concept of sparse tensors with an undefined fill value will make the sparse tensor a more general concept than dense tensor.

The proposal "a sparse tensor is a drop-in replacement of dense tensors for all tensor operations" has the following meaning: any function that implements some mathematical algorithm for dense tensors ought to able to handle sparse tensors with defined fill-values as inputs due to the one-to-one correspondence mentioned above. So, the contradiction you noted can be fixed by making the proposal more accurate: "a sparse tensor with defined fill-value is a drop-in replacement of ...".

Note that the undefined fill-value behavior, which is to ignore unspecified entries, already exists in PyTorch. For instance, sparse softmax ignores the contribution from unspecified tensor elements. Although, this can be conveniently modeled by defining `-inf` as fill-value because of softmax mathematical definition.
I imagine that the undefined fill-value feature would be also useful in manipulating sparse tensors representing weighted adjacency matrices from the graph theory or in other applications where the sparsity structure must be preserved when applying a function to sparse tensor elements. For instance, the undefined fill-value would allow distinguishing the following use cases:
```
x = [*, 1]                   # `*` denotes unspecified element
y1 = exp(x) -> [*, exp(1)]   # x fill-value is undefined, y1 fill-value is undefined
y2 = exp(x) -> [*, exp(1)]   # x fill-value is 0, y2 fill-value is 1
sum(y1) != sum(y2)
```

Finally, because sparse tensors with undefined fill-value do not represent mathematical tensors (objects that represent elements from the set `\mathbb{R}^{m \times n \times \cdots}`), many linear algebra matrix operations and algorithms are not defined for such sparse tensors (because operations between defined and undefined values are not defined).

rgommers:
Let me also add this related to undefined fill value, from the scikit-learn docs:

        Two semantics for sparse matrices are used in Scikit-learn:

        matrix semantics
            The sparse matrix is interpreted as an array with implicit and
            explicit zeros being interpreted as the number 0.  This is the
            interpretation most often adopted, e.g. when sparse matrices
            are used for feature matrices or :term:`multilabel indicator
            matrices`.
        graph semantics
            As with :mod:`scipy.sparse.csgraph`, explicit zeros are
            interpreted as the number 0, but implicit zeros indicate a masked
            or absent value, such as the absence of an edge between two
            vertices of a graph, where an explicit value indicates an edge's
            weight. This interpretation is adopted to represent connectivity
            in clustering, in representations of nearest neighborhoods
            (e.g. :func:`neighbors.kneighbors_graph`), and for precomputed
            distance representation where only distances in the neighborhood
            of each point are required.

RFC0003-sparse-roadmap/README.md
```diff
@@ -0,0 +1,1352 @@
...
+  Compressed Row/Column Storage format (CRS/CCS).
+
+The original version of GCRS/GCCS format uses a specific choice of
+dimensionality reduction where the dimensions are split into two parts
+using the evenness of dimensions numbers. This choice does not seem to
+have obvious advantages over other possible choises from the view
+point of data locality and performance. So, our modification will be
+parameterized verion of GCRS/GCCS where the choice of splitting the
+set of dimensions to two sets can be user's choice. Also, this
+approach allows one implementation for both CRS and CCS format cases.
```
ezyang, line 1084: @v0dro Parameterizing GCRS/GCCS doesn't seem like an unambiguous win. In particular, whenever you parametrize in this way, you have to record extra metadata regarding what parametrization was chosen. If there is a fixed convention (like the original paper's even/odd), you no longer need to store this data. Similarly, it's not obvious to me that one implementation supporting both CRS and CCS is actually a good thing! It certainly makes life a lot harder for prospective kernel writers.

I am a little uncertain whether or not parametrized form makes the indexing formulas more complicated or not because arbitrary rank indexing is difficult no matter what you do (since it's a loop, not a fixed formula).

RFC0003-sparse-roadmap/README.md
```diff
@@ -0,0 +1,1352 @@
...
+sparse tensor format to store the data may lead to more performant
+programs due to the efficient uses of memory as well as computational
+resources.
+
+The starting point of this proposal and the far-reaching goal is that
+the PyTorch tensor represents a multi-dimensional array whereas the
+used data storage format is an implementation detail that users can
+select depending on their data to (indirectly) control the choise of
+computational tools that would be most efficient for users'
+applications. Switching a storage format of data to another data
```
wenleix, line 67: If I understand correctly, it's a non-goal to support some method very specific to sparse tensors, such as preconditioned iterative methods? Because there might be more stuff specific to sparse tensor (which may not be even applicable to dense tensor)-- such how to partitioning the data, choosing the right preconditioner and Krylov subspace method, etc?

wenleix:
@pearu, @ezyang  

> I imagine that the undefined fill-value feature would be also useful in manipulating sparse tensors representing weighted adjacency matrices from the graph theory or in other applications where the sparsity structure must be preserved when applying a function to sparse tensor elements. For instance, the undefined fill-value would allow distinguishing the following use cases:

Note graph applications can also be modeled (and I personally found with clearer semantic) as an adjacent matrix  **AND** a weight matrix. 

Another way to view sparse tensor with underfined fill-value as  a "compressed encoding" to both the adjacent matrix + weight matrix with one "sparse Tensor" -- the corresponding dense tensor, if we have to define it, has to be two tensors instead of 1 tensor.

PR #8, opened 2020-09-13
PyTorch Sparse Tensors: fill-value property

pearu:
As in title.

### DIFF
--- a/RFC-0004-sparse-fill-value/README.md
+++ b/RFC-0004-sparse-fill-value/README.md
@@ -1,0 +1,509 @@
+# PyTorch Sparse Tensors: fill-value property
+
+|            |                 |
+| ---------- | --------------- |
+| Authors    | Pearu Peterson  |
+| Status     | Draft           |
+| Type       | Process         |
+| Created    | 2020-09-08      |
+| Resolution | TBD             |
+
+## Abstract
+
+This proposal introduces a fill-value property to PyTorch sparse
+tensors that generalizes the current interpretation of unspecified
+elements from zero value to any value, including undefined value as a
+future extension.
+
+## Motivation and Scope
+
+In general, the unspecified elements of sparse tensors have
+domain-specific interpretations:
+- In Linear Algebra domain, the unspecified elements are zero valued.
+- In Graph domain, the unspecified elements of adjacency matrices
+  represent non-edges. In this document, a non-edge corresponds to
+  undefined value.
+- In neural networks, sparse tensors can be inputs, say, to activation
+  functions that are defined in terms of elementary functions.
+  Evaluation of such functions on sparse tensors can be element-wise,
+  or one-dimension-wise operations. This includes the evaluation of
+  the functions on the values corresponding to unspecified
+  elements. Even when the unspecified elements are initially being
+  zero valued, the values of unspecified elements may be mapped to
+  non-zero values. So, in the domain of Calculus, the unspecified
+  elements of sparse tensors can have any defined value.
+
+Currently, in PyTorch, the unspecified elements of sparse tensors are
+zero valued, but not only. For instance, the `torch.sparse.softmax`
+function assumes that the unspecified elements of a sparse tensor
+input are negative infinity (-inf) valued.
+
+In PyTorch 1.6, element-wise functions from Calculus such as `exp`,
+`log`, etc, or arithmetic operations such as addition, subtraction,
+etc, on sparse tensors are not supported, because the existing
+functions on sparse tensors assume that the unspecified elements are
+zero valued (this applies to functions in `torch` namespace, functions
+in `torch.sparse` namespace may use different interpretation for
+unspecified elements).
+
+To support Calculus functions on sparse tensors, we'll propose adding a
+fill-value property to PyTorch sparse tensors that will represent the
+values of unspecified sparse tensor elements.  While doing so, we also
+need to consider
+- how the non-zero fill-value affects the result of linear algebra
+  operations such as matrix multiplication, etc.
+- how to define a fill-value for hybrid tensors (tensors with sparse
+  and dense dimensions),
+- how PyTorch autograd support needs to deal with a defined
+  fill-value.
+
+
+## Proposal
+
+This proposal is about enabling sparse tensor support for Calculus
+domain while preserving the existing functionality for Linear Algebra
+domain, and allowing extensions to Graph domain.
+
+In the following, we describe the sparse tensor fill-value feature
+with examples to sparse tensors in COO storage format.  When a new
+sparse tensor storage format is introduced to PyTorch, the same
+semantics apply to the new format.
+
+1.  We propose to extend sparse tensor constructors with a keyword
+    argument `fill_value` that users can use to define the value for
+    unspecified elements of the constructed sparse tensor.
+
+    For instance, the Python signature of `sparse_coo_tensor` would be
+
+    ```python
+    torch.sparse_coo_tensor(indices, values, size=None, fill_value=None, dtype=None, device=None, requires_grad=False)
+    ```
+
+2.  The default fill-value is `0`.
+
+    This choice is consistent with the interpretation of unspecified
+    element in Linear Algebra domain.
+
+3.  PyTorch functions that have `layout` argument, may use different
+    fill-value when constructing a sparse tensor as defined by the
+    following table:
+
+    | Function              | Fill-value of returned sparse tensor |
+    | :-------------------- | :----------------------------------- |
+    | `torch.arange`        | 0                                    |
+    | `torch.empty`         | 0                                    |
+    | `torch.empty_like`    | 0                                    |
+    | `torch.empty_strided` | N/A                                  |
+    | `torch.eye`           | 0                                    |
+    | `torch.full`          | same as `fill_value` argument        |
+    | `torch.full_like`     | same as `fill_value` argument        |
+    | `torch.linspace`      | 0                                    |
+    | `torch.logspace`      | 0                                    |
+    | `torch.ones`          | 1                                    |
+    | `torch.ones_like`     | 1                                    |
+    | `torch.range`         | 0                                    |
+    | `torch.zeros`         | 0                                    |
+    | `torch.zeros_like`    | 0                                    |
+
+    For discussion: should we introduce the `fill_value` argument to
+    functions `empty` and `empty_like`?
+
+4.  The fill-value of a sparse tensor can be acquired via
+    `fill_value()` method that returns a strided `torch.Tensor`
+    instance that has the same dtype and storage location properties
+    as the elements of the sparse tensor:
+
+    - `A.fill_value().dtype == A.dtype`
+    - `A.fill_value().device == A.device`
+
+    The fill-value of a sparse non-hybrid tensor is a scalar tensor
+    `torch.tensor(0, dtype=A.dtype, device=A.device)`.
+
+5.  The fill-value of a hybrid sparse tensor has the same shape as the
+    dense part of the tensor.
+
+    For instance, for a sparse tensor `A` in COO storage format we
+    have:
+
+    - `A.fill_value().shape == A.values().shape[1:]`
+
+    While an obvious alternative to tensor fill-value would be a
+    scalar fill-value, the scalar fill-value would complicate the
+    implementation of algorithms for hybrid tensors. For
+    instance:
+
+    1. Accessing a dense element of a hybrid sparse tensor, the result
+       is a dense tensor. When this element corresponds to unspecified
+       element, one needs to construct a new tensor and fill it with
+       the scalar fill-value. A more efficient and simple way is to
+       just return the output of `fill_value()` that does would not
+       involve constructing a new tensor on each method call.
+
+    2. When applying a row-wise function (such as `softmax`, for
+       instance) to a hybrid tensor, each row may induce a different
+       fill-value.
+
+       For example, consider a 2D hybrid tensor with one sparse and
+       one dense dimension in COO format, and apply softmax along the
+       first dimension assuming the default fill-value 0:
+
+       ```python
+       >>> A = torch.sparse_coo_tensor(indices=[[0, 3]],
+                                       values=[[.11, .12], [.31, .32]],
+                                       size=(4, 2))
+       >>> A.to_dense()
+       tensor([[0.1100, 0.1200],
+               [0.0000, 0.0000],
+               [0.0000, 0.0000],
+               [0.3100, 0.3200]])
+       >>> A.to_dense().softmax(0)
+       tensor([[0.2492, 0.2503],
+               [0.2232, 0.2220],
+               [0.2232, 0.2220],
+               [0.3044, 0.3057]])
+       ```
+
+       Observe that the softmax result can be represented as a sparse
+       tensor with the same sparsity structure (same indices) as the
+       input only when the fill-value is a 1-D dense tensor, that is,
+
+       ```python
+       A.softmax(0) == torch.sparse_coo_tensor(indices=A.indices(),
+                                               values=[[0.2492, 0.2503], [0.3044, 0.3057]],
+                                               size=A.size(),
+                                               fill_value=[0.2232, 0.2220])
+       ```
+
+       If fill-value would be defined as scalar, the result of
+       `A.softmax(0)` would need to be a dense tensor or raise
+       exception about unsupported feature.
+
+6.  The fill-value of a sparse tensor can be changed in-place.
+
+    For instance,
+
+    ```python
+    A.fill_value().fill_(1.2)
+    ```
+
+    resets the fill value of a non-hybrid sparse tensor `A` to `1.2`.
+
+7.  If `A` is a sparse tensor and `f` is any calculus function that is
+    applied to a tensor element-wise, then
+
+    ```python
+    f(A) == torch.sparse_coo_tensor(A.indices(), f(A.values()), fill_value=f(A.fill_value()))
+    ```
+
+    Note that if `A` would be using COO storage format then this
+    relation holds only if `A` is coalesced (`A.values()` would throw
+    an exception otherwise).
+
+8.  The fill-value of an element-wise n-ary operation on sparse
+    tensors with different fill-values is equal to the result of
+    applying the operation to the fill-values of the sparse
+    tensors.
+
+    For instance (`*` represents unspecified element),
+
+    ```
+    A = [[1, *], [3, 4]]                  # A fill-value is 2
+    B = [[*, 6], [*, 8]]                  # B fill-value is 7
+    A + B = [[1 + 4, 2 + 4], [*, 6 + 8]]  # A + B fill-value is 2 + 7 = 9
+    ```
+
+9.  Existing PyTorch functions that support sparse tensor as inputs,
+    need to be updated for handling the defined fill-values. This will
+    be implemented in two stages:
+
+    1. All relevant functions need to check for zero fill-value. If a
+       non-zero fill-value is used, raise an exception.
+
+       For instance:
+
+       ```C++
+       TORCH_CHECK(A.fill_value().nonzero().numel() == 0,
+                   "The <function> requires a sparse tensor with zero fill-value, got ",
+                   A.fill_value());
+       ```
+
+       This check ensures that the existing functions will not produce
+       silently incorrect results when user passes a sparse tensor
+       with non-zero fill-value as an input while the function assumes
+       it to be zero.
+
+       For discussion: Since the zero fill-value check is required in
+       all functions supporting sparse inputs, we may need a more
+       efficient way than `.nonzero().numel() == 0` to determine if a
+       tensor contains only zeros. This efficiency issue is relevant
+       only for hybrid tensors with a large dense part.
+
+    2. Update the related functions to handle non-zero fill-values of
+       input sparse tensors correctly.
+
+       For instance, consider a matrix multiplication of two sparse
+       tensors `A` and `B` with fill-values `a` and `b`, respectively,
+       then the `matmul` operation can be expanded as follows:
+
+       ```python
+       matmul(A, B) = matmul(A - fA + fA, B - fB + fB)
+                    = matmul(A - fA, B - fB) + fA * matmul(ones_like(A), B) + fB * matmul(A, ones_like(B))
+       ```
+
+       where the first term can be computed using existing matmul for
+       sparse tensors with zero fill-value, and the last two terms can
+       be replaced with a computation of a single row or column of the
+       corresponding matrix products that has reduced computational
+       complexity.
+
+10. Sparse tensors with defined fill-value have intrinsic constraints
+    between all the unspecified tensor elements (these are always
+    equal) that must be taken into account when implementing Autograd
+    backward methods for functions that receive sparse tensors as
+    inputs.
+
+    Sparse tensors with undefined fill-value don't have the intrinsic
+    constraints as discussed above.
+
+
+### Future extensions and existing issues
+
+11. For Graph domain, the undefined fill-value can be specified as a
+    tensor with zero dimension(s) that satisfies all the relations
+    listed above except point 5. Invalidation of the point 5 will
+    provide a consistent way to differentiate between defined and
+    undefined fill-values.
+
+    For instance, to reset a fill-value of a sparse tensor to undefined
+    fill-value, one can use:
+
+    ```python
+    A.fill_value().resize_((0,) * len(A.values().shape[1:]))
+    ```
+
+    Note that this operation is reversible, that is, to reset a
+    undefined fill-value to a defined value, say `1.2`, one can use:
+
+    ```python
+    A.fill_value().resize_(A.values().shape[1:]).fill_(1.2)
+    ```
+
+    This works only when the initial fill-value of `A` is a defined
+    value with the shape constraints specified in point 5.
+
+12. For Statistics domain, where unspecified elements are interpreted
+    as "missing data". The missing data concept is different from
+    undefined value in the sense that it represents a certain
+    measurement flaw that cannot be foreseen to appear in a data while
+    the undefined value represents a structural lack of data that is
+    determined by the system under study.
+
+    The fill-value corresponding to missing data can be specified as
+    `nan` (or a tensor of `nan`-s in case of hybrid tensor). This
+    choice is consistent with convention used in Python Pandas
+    package.
+
+13. The introduction of non-zero fill-value feature requires
+    revisiting the existing sparse tensor API.
+
+    1. The acronym NNZ means the "Number of Non-Zeros" in a sparse
+       tensor. In PyTorch, this acronym is used in several places:
+
+       - in the implementations of sparse tensor and related
+         functionality,
+       - in the `repr` output of COO sparse tensor, see
+         [here](https://pytorch.org/docs/master/generated/torch.sparse_coo_tensor.html#torch.sparse_coo_tensor),
+       - as a private method `_nnz()`, see
+         [here](https://pytorch.org/docs/master/sparse.html?highlight=nnz#torch.sparse.FloatTensor._nnz),
+       - as a optional keyword argument `check_sparse_nnz` in
+         [torch.autograd.gradcheck](https://pytorch.org/docs/master/autograd.html#numerical-gradient-checking).
+
+       The acronym NNZ is misused in PyTorch:
+
+       - `nnz` holds the value of the "Number of Specified Elements"
+         (NSE) in a sparse tensor
+       - `nnz` is not always equal to the number of zeros in the
+         sparse tensor, for instance, the `values` of the sparse
+         tensor in COO format may contain zero values that are not
+         accounted in `nnz`
+
+       With the introduction of non-zero fill-value, the misuse of
+       acronym NNZ will deepen because with non-zero fill-value the
+       sparse tensor may have no zero elements, e.g. `torch.full((10,
+       10), 1.0, layout=torch.sparse_coo)` for which `nnz` would be
+       `0`.
+
+       Possible actions:
+
+       - Stop the misuse of NNZ acronym: (i) replace the usage of
+         "NNZ" with "NSE", (ii) deprecate the use of `_nnz()` in favor
+         or `_nse()`, (iii) remove `_nnz()` starting from PyTorch 2.0.
+       - Do nothing.  This is the (undocumented) approach taken in
+         Wolfram Language where [one can use "NonzeroValues" to
+         determine the number of specified
+         elements](https://community.wolfram.com/groups/-/m/t/1168496)
+         even when the fill-value specified is non-zero.
+       - Other ideas?
+
+14. This proposal is relevant to the following PyTorch issues:
+
+   - [Sparse tensor eliminate zeros](https://github.com/pytorch/pytorch/issues/31742)
+   - [Add sparse softmax/log_softmax functionality (ignore zero entries)](https://github.com/pytorch/pytorch/issues/23651)
+   - [The state of sparse tensors](https://github.com/pytorch/pytorch/issues/9674)
+
+## Application: random sequence of rare events with non-zero mean
+
+Let's assume one has a device that periodically outputs a
+mostly-constant signal with some occasional spikes, for instance, from
+the neuroscience measurements of cells membrane potential:
+
+<a href="https://commons.wikimedia.org/wiki/File:IPSPsummation.JPG"><img src="https://upload.wikimedia.org/wikipedia/commons/a/a1/IPSPsummation.JPG" width="50%"/></a>
+
+
+For simplicity, let the corresponding signal be
+
+```
+period:  0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 ... 1000000
+signal:  5  5  5  7  5  5  5  5  6  8  5  5  5  5  5  5  5  9  5 ... 5
+```
+
+This sequence can be stored efficiently in a sparse tensor with `5` as
+fill-value:
+
+```python
+signal = torch.sparse_coo_tensor(indices=[[3, 8, 9, 17, ...]],
+                                 values=  [7, 6, 8,  9, ...],
+                                 size=(1000001,),
+                                 fill_value=5)
+```
+
+One can conveniently convert the signal to the relevant physical unit,
+say, to [membrane
+potential](https://en.wikipedia.org/wiki/Membrane_potential), using a
+scalar conversion factor:
+
+```python
+mV_unit = -8.0
+membrane_potential = signal * mV_unit
+```
+
+that can be then used in computing other physical quantities, for instance,
+
+```python
+intracellular_concentration_of_potassium = extracellular_concentration_of_potassium * torch.exp(-a * membrane_potential)
+```
+
+where `a` is a constant defined by [Nernst
+equation](https://en.wikipedia.org/wiki/Nernst_equation). Notice that
+the result `intracellular_concentration_of_potassium` is a sparse
+tensor with a non-zero fill-value.
+
+Without the non-zero fill-value support, the researchers of the field
+would not be able to compute these quantities as simply as above
+because they would need to keep track on the constant level of the
+signals at different transformation steps. For instance, using Pytorch
+1.6, the same computation of intracellular concentration of potassium
+would be:
+
+```python
+# for membrane_potential:
+membrane_potential_base = 5 * mV_unit
+membrane_potential_var_values = signal.values() * mV_unit
+# for torch.exp(a * membrane_potential):
+factor_base = torch.exp(-a * membrane_potential_base)
+factor_var_values = torch.exp(-a * membrane_potential_var_values)
+# for intracellular_concentration_of_potassium:
+intracellular_concentration_of_potassium_base = extracellular_concentration_of_potassium * factor_base
+intracellular_concentration_of_potassium_var_values = intracellular_concentration_of_potassium * factor_var_values
+
+intracellular_concentration_of_potassium_var = torch.sparse_coo_tensor(
+    indices = signal.indices(),
+    values = intracellular_concentration_of_potassium_var_values,
+    size = signal.size())
+```
+
+which is convoluted, error-prone, and does not lead to solitary data
+structure as is `intracellular_concentration_of_potassium` above.
+
+### Review of sparse array software
+
+As a rule, software that implement sparse matrix support, do not
+implement support for non-zero fill-value. The reason for this is
+two-fold: either the application of the software do not require this
+feature (such as applications from Linear Algebra and Graph domains),
+or the feature declared not implemented yet. However, there exists
+also software that implement the non-zero fill-value support for
+sparse matrices.  In the following, we will review a selection of
+sparse array software to learn the different approaches how the
+fill-value feature (or the lack of it) is handled.
+
+### Wolfram Language
+
+[SparseArray](https://reference.wolfram.com/language/ref/SparseArray.html)
+has a constructor variant `SparseArray[data,dims,val]` that "yields a
+sparse array in which unspecified elements are taken to have value
+`val`". The default fill-value is 0.
+
+### Pandas - Python Data Analysis Library
+
+Pandas
+[SparseArray](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.arrays.SparseArray.html#pandas.arrays.SparseArray)
+supports `fill_value` option to enable memory saving for most common
+value in an array. The default fill-value depends on array dtype, for
+instance, it is `nan` for `float` and `0` for `int`. The
+interpretation of the default fill-value is "missing data".
+
+### MathWorks
+
+Matlab's [sparse
+matrix](https://www.mathworks.com/help/matlab/ref/sparse.html) does
+not support non-zero fill-value feature. Historically, the idea of
+non-zero fill-value was rejected in the interest of simplicity, [see
+Sec 2.5 in this
+paper](https://epubs.siam.org/doi/abs/10.1137/0613024).
+
+### Scikit-learn/Scipy.sparse
+
+[Scikit-learn defines two
+semantics](https://scikit-learn.org/stable/glossary.html#term-sparse-matrix)
+for interpreting the unspecified entries in a sparse matrix:
+
+- Matrix (here Linear Algebra) semantics where the fill-value is
+  assumed to be 0,
+- Graph semantics where the fill-value is assumed to be undefined
+  value.
+
+The scikit-learn algorithms of the Matrix and Graph domains use sparse
+arrays from
+[scipy.sparse](https://docs.scipy.org/doc/scipy/reference/sparse.html)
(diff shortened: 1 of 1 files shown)


RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,509 @@
...
+       ```python
+       A.softmax(0) == torch.sparse_coo_tensor(indices=A.indices(),
+                                               values=[[0.2492, 0.2503], [0.3044, 0.3057]],
+                                               size=A.size(),
+                                               fill_value=[0.2232, 0.2220])
+       ```
+
+       If fill-value would be defined as scalar, the result of
+       `A.softmax(0)` would need to be a dense tensor or raise
+       exception about unsupported feature.
```
rgommers, line 179: `sparse.softmax` works today for this case right? And returns a dense tensor? I think that cannot be changed after release anymore.

rgommers: This again seems a trade-off: for new functions that gain sparse support and are `softmax`-like, you can return a sparse rather than a dense tensor for hybrid input (potentially saving memory) - but you're paying with increased memory use of the input tensor as well - even in use cases that never need the non-scalar fill value feature.

In addition, it makes implementation of algorithms more complex, and the interface as well (see, e.g., point 6 below which now needs a disclaimer about `.fill()` only working for non-hybrid tensors).

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,509 @@
...
+    as "missing data". The missing data concept is different from
+    undefined value in the sense that it represents a certain
+    measurement flaw that cannot be foreseen to appear in a data while
+    the undefined value represents a structural lack of data that is
+    determined by the system under study.
+
+    The fill-value corresponding to missing data can be specified as
+    `nan` (or a tensor of `nan`-s in case of hybrid tensor). This
+    choice is consistent with convention used in Python Pandas
+    package.
```
rgommers, line 304: This choice is problematic, and Pandas is introducing its own `NA` support instead of using `nan`. The reason is that it doesn't work for integers, because there is no `nan`.

If fill value was scalar, this would be much easier, use an empty tensor for example.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,509 @@
...
+         [torch.autograd.gradcheck](https://pytorch.org/docs/master/autograd.html#numerical-gradient-checking).
+
+       The acronym NNZ is misused in PyTorch:
+
+       - `nnz` holds the value of the "Number of Specified Elements"
+         (NSE) in a sparse tensor
+       - `nnz` is not always equal to the number of zeros in the
+         sparse tensor, for instance, the `values` of the sparse
+         tensor in COO format may contain zero values that are not
+         accounted in `nnz`
```
rgommers, line 328: This is a good point. +1 for `nse` rather than `nnz`.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,509 @@
...
+    | `torch.linspace`      | 0                                    |
+    | `torch.logspace`      | 0                                    |
+    | `torch.ones`          | 1                                    |
+    | `torch.ones_like`     | 1                                    |
+    | `torch.range`         | 0                                    |
+    | `torch.zeros`         | 0                                    |
+    | `torch.zeros_like`    | 0                                    |
+
+    For discussion: should we introduce the `fill_value` argument to
+    functions `empty` and `empty_like`?
```
rgommers, line 109: I'd prefer not doing that, given that it'd be a keyword that does not apply to dense tensors. One can easily set the fill value after construction.

Also, `empty` gives undefined values by definition, it just reserves a block of memory, so having already-defined fill values seems not needed.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,509 @@
...
+  undefined value.
+- In neural networks, sparse tensors can be inputs, say, to activation
+  functions that are defined in terms of elementary functions.
+  Evaluation of such functions on sparse tensors can be element-wise,
+  or one-dimension-wise operations. This includes the evaluation of
+  the functions on the values corresponding to unspecified
+  elements. Even when the unspecified elements are initially being
+  zero valued, the values of unspecified elements may be mapped to
+  non-zero values. So, in the domain of Calculus, the unspecified
+  elements of sparse tensors can have any defined value.
```
rgommers, line 34: The neuroscience use case towards the end of this RFC is a good motivating use case. It would be useful to link to it from this section.

rgommers (commented):
Looks great, thanks @pearu!

rgommers:
@mruberry this document has the different interpretations of sparse tensors that we were talking about.

RFC-0004-sparse-fill-value/README.md
pearu: I now agree on not introducing `fill_value` to `empty` because the tensor content is arbitrary (in the sense that the content is not initialized), hence the `fill_value` content should be also arbitrary.

Re "Also, empty gives undefined values by definition...": I think by "undefined value" you mean "arbitrary value" as defined above. In this proposal, we are using the term "undefined fill-value" that has a different meaning: undefined fill-value is a not-a-value, like `None` in Python is sometimes used to represent the situation when a value has not been specified.

That said, I think the proposal should use "unspecified fill-value" (any better idea?) instead of "undefined fill-value" which is too similar to "undefined value" that has a different meaning: https://en.wikipedia.org/wiki/Undefined_value

rgommers: > I think by "undefined value" you mean "arbitrary value" as defined above.

Yes indeed. 

> That said, I think the proposal should use "unspecified fill-value" (any better idea?) instead of "undefined fill-value" 

Agreed, sounds like the clearest terminology to me.

RFC-0004-sparse-fill-value/README.md
pearu: Good idea. Done.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,509 @@
...
+    scalar fill-value, the scalar fill-value would complicate the
+    implementation of algorithms for hybrid tensors. For
+    instance:
+
+    1. Accessing a dense element of a hybrid sparse tensor, the result
+       is a dense tensor. When this element corresponds to unspecified
+       element, one needs to construct a new tensor and fill it with
+       the scalar fill-value. A more efficient and simple way is to
+       just return the output of `fill_value()` that does would not
+       involve constructing a new tensor on each method call.
```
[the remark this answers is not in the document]
pearu: The "more memory" argument is not really a problem considering that the `fill_value` would consume the same amount of memory that a single element of a sparse tensor does. I think we can assume that in a typical application of sparse tensors, the number of specified elements is far greater than 1.

For example, consider a 1000x1000 matrix with 10000 specified elements. Optimizing for memory would give saving 1/10000 of total memory usage for this particular example. However, if one would loop, say, over all elements of the tensor, there would be 1000000 minus 10000 newly created tensor instances corresponding to all unspecified elements. While the second example is somewhat extreme, I would give more weight to not creating a bunch of new tensor instances than saving 0.01 % of memory.

RFC-0004-sparse-fill-value/README.md
pearu: > `sparse.softmax` works today for this case right? And returns a dense tensor? I think that cannot be changed after release anymore.

No. `torch.sparse.softmax` always returns a sparse tensor because the assumed fill-value is `-inf`.
`torch.softmax` does not support sparse tensors (because of the lack of fill-value).

pearu: To be clear: "dense tensor" can use different storage format layouts. Typically it is `strided` but it could be also `sparse_coo` with `nnz==product(shape)`.

pearu: > This again seems a trade-off: for new functions that gain sparse support and are `softmax`-like, you can return a sparse rather than a dense tensor for hybrid input (potentially saving memory) - but you're paying with increased memory use of the input tensor as well - even in use cases that never need the non-scalar fill value feature.

See https://github.com/pytorch/rfcs/pull/8#discussion_r487567061 that basically eliminates the trade-off concern.

pearu: > In addition, it makes implementation of algorithms more complex, and the interface as well (see, e.g., point 6 below which now needs a disclaimer about `.fill()` only working for non-hybrid tensors).

I disagree. If the fill-value can be only a scalar, the result of `softmax` on sparse tensor is always dense (in the sense that all tensor elements are specified). The memory trade-off of this will be huge compared to the memory trade-off of having non-scalar fill-value for hybrid tensors.

The example in point 6. does not need the non-hybrid disclaimer, it would work also for hybrid tensors.

In general, when applying an operation to a tensor, the operation should result in a new tensor with the same storage format as the input tensor (unless the operation explicitly advertises or provides via `layout` or `out` argument a different behavior). This design pattern is used ,say, in Matlab, and IMHO this is a good pattern to follow.

RFC-0004-sparse-fill-value/README.md
pearu: I agree, `nan` is a bad choice, in general. PyTorch currently supports only float sparse tensors so the lack-of-nan-for-an-integer does not apply, at least, not currently.

The shape of fill-value (scalar or no scalar) is irrelevant, empty tensors can have arbitrary shapes provided that at least one dimension size is 0.

That said, I now think that I got the usage of sparse tensors in Statistics domain wrong. "missing data" exist as specified values (be them `nan`, `N/A`, etc), the fill-value represents elements havng the most common value. I'll fix that..

pearu: A reasonable fix seems to be removing the discussion of Statistics domain from the document.

RFC-0004-sparse-fill-value/README.md
pearu: Second thought: " the unspecified fill-value can be specified..." sounds strange. So, a new suggestion is to use "indefinite fill-value" or "indeterminate fill-value".

pearu: Using "indefinite fill-value".

RFC-0004-sparse-fill-value/README.md
pearu: Removed extension to Statistics domain as not needed.

RFC-0004-sparse-fill-value/README.md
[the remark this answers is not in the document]
rgommers: I'm thinking there are sparse tensors that are consisting of _only_ dense blocks (e.g. to represent block-sparse type structure), and for those types of tensors memory use would double.

Am I misunderstanding something here about the hybrid format?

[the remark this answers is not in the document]
rgommers: Oh wait, okay - my assumption here about what you were proposing was wrong. Agreed, this seems like a very minor concern.

RFC-0004-sparse-fill-value/README.md
rgommers: > In general, when applying an operation to a tensor, the operation should result in a new tensor with the same storage format as the input tensor (

That makes sense as a principle, agreed.

RFC-0004-sparse-fill-value/README.md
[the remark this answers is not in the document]
pearu: PyTorch hybrid tensor is a sparse tensor in COO format, that is, it has `indices` and `values` to store the tensor data, with the difference that the size of `values` elements is not 0 (that would be a scalar and correspond to traditional COO tensor) but can have any dimension. `values` is a strided tensor. See the invariants in https://pytorch.org/docs/stable/sparse.html

So, no, hybrid tensor does not represent a block-sparse type of matrix where the matrix consists of rectangular blocks that are dense and anything in between are zeros. Hybrid tensor can represent certain types of block-sparse matrices: the blocks would need to be full rows, for instance.

Introducing tensor fill-value to hybrid tensors would certainly not double the memory usage. The increase of memory usage would be equal to the memory usage of one element of the sparse part of the hybrid tensor.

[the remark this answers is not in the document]
pearu: See https://ars.els-cdn.com/content/image/1-s2.0-S0743731518305161-gr1.jpg where
- the first image represents a traditional 3D sparse tensor
- the second image represents a 3D hybrid tensor with 2 sparse dimensions and 1 dense dimension.
- the last image represents a 3D tensor with 1 sparse dimension and 2 dense dimensions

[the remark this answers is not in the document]
hameerabbasi: One can achieve the benefits of having both a dense and a scalar by saying simply that the fill-value should be broadcast-able to the dense shape of the tensor, instead of restricting it to just be the dense shape.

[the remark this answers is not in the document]
pearu: @hameerabbasi I like this idea a lot because it would resolve the zero fill-value test performance issue.
For element-wise operations, the idea works. For row-wise operations (like softmax), the idea needs to be analyzed because there the fill-value would need to be resized from scalar to 1D tensor in-place.

I'll unresolve the conversation to address @hameerabbasi point only.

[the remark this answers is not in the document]
hameerabbasi: Yes, in the case of softmax it would need a new allocation, or else you would need a way of looping over strided fill-values within sparse kernels. The latter would also resolve your issue.

RFC-0004-sparse-fill-value/README.md
pearu: Made the rename nnz->nse as a recommendation.

RFC-0004-sparse-fill-value/README.md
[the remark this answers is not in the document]
pearu: According to updated proposal, one can specify hybrid tensor fill value that shape can be smaller that the shape of tensor dense elements. The specified fill value is available via `_fill_value()` method while the fill value of hybrid tensor can be acquired via `fill_value()`. Element-wise ops are applied to `_fill_value()` and this is also used for zero fill value testing. This approach resolves the performance issue optimally.

RFC-0004-sparse-fill-value/README.md
pearu: I'll close this conversation as resolved as it seems to originate from a misunderstanding of hybrid tensors. Feel feel to reopen.

pearu:
@rgommers @hameerabbasi, please review. The updated proposal resolves many open-ended questions we had previously and the proposal should be complete now.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+    equal to `<constructed sparse tensor>.fill_value().shape`.
+
+12. Sparse tensors with defined fill value have intrinsic constraints
+    between all the unspecified tensor elements (these are always
+    equal) that must be taken into account when implementing Autograd
+    backward methods for functions that receive sparse tensors as
+    inputs.
+
+    Sparse tensors with indefinite fill value don't have the intrinsic
+    constraints as discussed above.
```
rgommers, line 347: Does autograd even make sense in that case? I suspect not, in which case just deleting this sentence would be good.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+       taken in Wolfram Language where [one can use "NonzeroValues" to
+       determine the number of specified
+       elements](https://community.wolfram.com/groups/-/m/t/1168496)
+       even when the fill value specified is nonzero.
+
+    2. The `torch` namespace functions `arange`, `range`, `linspace`,
+       and `logspace` have `layout` argument that is not needed.
+
+       Currently, PyTorch defines three layouts: `strided`,
+       `sparce_coo`, and `_mkldnn`. Because the mentioned functions
```
rgommers, line 409: typo in `sparce_coo`

rgommers (commented):
LGTM, just a few small comments.

You can remove the "Draft" status I think.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+                                               size=A.size(),
+                                               fill_value=[0.2232, 0.2220])
+       ```
+
+       The alternative, if the fill value was defined as scalar,
+       would be memory inefficient because the result of
+       `A.softmax(0)` is a full sparse tensor with all elements
+       specified.
+
+6.  The fill value specified that is specified in the `fill_value`
```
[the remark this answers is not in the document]
pearu: fixed

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+8.  If `A` is a sparse tensor and `f` is any calculus function that is
+    applied to a tensor element-wise, then:
+
+    ```python
+    f(A) == torch.sparse_coo_tensor(A.indices(), f(A.values()), fill_value=f(A._fill_value()))
+    ```
+
+    Note that if `A` would be using COO storage format then this
+    relation holds only if `A` is coalesced (`A.values()` would throw
+    an exception otherwise).
```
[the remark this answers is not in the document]
pearu: I agree, it was surprising also for me. On the other hand, if sparse tensors will implement all the traditional features that strided tensors have (slicing, etc), then for end-users, `indices` and `values` become implementation detail, or for expert users only.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+    tensors with different fill values is equal to the result of
+    applying the operation to the fill values of the sparse
+    tensors.
+
+    For instance (`*` represents unspecified element),
+
+    ```
+    A = [[1, *], [3, 4]]                  # A fill value is 2
+    B = [[*, 6], [*, 8]]                  # B fill value is 7
+    A + B = [[1 + 3, 2 + 4], [*, 6 + 8]]  # A + B fill value is 2 + 7 = 9
```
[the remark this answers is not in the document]
pearu: Thanks for spotting this, fixed.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
+# PyTorch Sparse Tensors: `fill_value` property
```
hameerabbasi, line 1: One can think of titles more descriptive to people not familiar with the topic at hand.
```suggestion
# Expanding PyTorch sparse tensor support with fill values
```

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+
+In PyTorch 1.6, element-wise functions from Calculus such as `exp`,
+`log`, etc, or arithmetic operations such as addition, subtraction,
+etc, on sparse tensors are not supported, because the existing
+functions on sparse tensors assume that the unspecified elements are
+zero valued (this applies to functions in `torch` namespace, functions
+in `torch.sparse` namespace may use a different interpretation for
+unspecified elements).
+
+To support Calculus functions on sparse tensors, we propose adding a
```
hameerabbasi, line 51: Perhaps don't specify "Calculus" here, instead just say "to widen support for sparse tensors"
```suggestion
To widen support for sparse tensors to include more functions, we propose adding a
```

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+
+In the following, we describe the sparse tensor `fill_value` feature
+with examples involving sparse tensors in COO storage format.  When a
+new sparse tensor storage format is introduced to PyTorch, the same
+semantics should be applied to the new format.
+
+0.  Used terminology
+
+    - "Defined value" is a value for which memory is allocated and
+      initialized.
```
hameerabbasi, line 77:
```suggestion
      initialized to some defined value.
```

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+    - "Undefined value" is a value for which memory is allocated, but
+      the content of the memory can be arbitrary.
```
hameerabbasi, line 79:
```suggestion
      the content of the memory can be arbitrary, perhaps due to garbage values left-over
      from old computations or another process.
```

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+      with many equal elements.
+    - "Strided tensor format" is a process-efficient storage format
+      for general tensors.
+    - "Dense tensor" is a tensor with many non-equal elements. Dense
+      tensors can be stored in any tensor format, including a sparse
+      tensor format.
+    - "Dense part" is a strided tensor that is obtained by fixing the
+      indices of all sparse dimensions in a hybrid tensor.
+
+1.  We propose to extend sparse tensor constructors with a keyword
```
hameerabbasi, line 91:
```suggestion
1.  We propose to extend sparse tensor constructors with an extra keyword
```

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+
+3.  PyTorch functions that have the `layout` argument may use a different
+    `fill_value` when constructing a sparse tensor as defined by the
+    following table:
+
+    | Function              | `fill_value` of returned sparse tensor |
+    | :-------------------- | :----------------------------------- |
+    | `torch.empty`         | undefined value                      |
+    | `torch.empty_like`    | undefined value                      |
+    | `torch.empty_strided` | N/A                                  |
```
hameerabbasi, line 121: This function makes no sense for sparse tensors, it makes no sense to include it in this table.
```suggestion
```

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+    See point 6 below for how the output of `fill_value()` method is
+    computed.
+
+5.  The fill value of a hybrid sparse tensor has the same shape as the
+    dense part of the tensor.
+
+    For instance, for a sparse tensor `A` in COO storage format we
+    have:
+
+    - `A.fill_value().shape == A.values().shape[1:]`
```
hameerabbasi, line 153: For a sparse tensor (not hybrid), `A.fill_value().shape == ()` as it has no dense part. I imagine this applies more generally to hybrid tensors.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+       ```python
+       matmul(A, B) = matmul(A - fA + fA, B - fB + fB)
+                    = matmul(A - fA, B - fB) + fA * matmul(ones_like(A), B) + fB * matmul(A, ones_like(B))
+       ```
+
+       where the first term can be computed using existing matmul for
+       sparse tensors with zero fill value, and the last two terms can
+       be replaced with a computation of a single row or column of the
+       corresponding matrix products that has reduced computational
+       complexity.
```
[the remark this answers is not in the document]
pearu: Agreed. Added the following sentence:
```
       In general, updating all linear algebra functions to support
       nonzero fill value will be a notable effort and it is not the
       aim of this proposal to seek for it immediately.
```

RFC-0004-sparse-fill-value/README.md
pearu: I think it does if Graph models would need the autograd support. But it makes to delete this as the proposal does not tackle all the details that will be related to indefinite fill value.

RFC-0004-sparse-fill-value/README.md
pearu: fixed

RFC-0004-sparse-fill-value/README.md
pearu: deleted

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+                                size=(4, 2),
+                                fill_value = 1.2)
+    A.fill_value() -> torch.tensor([1.2, 1.2])
+    A._fill_value() -> torch.tensor(1.2)
+    ```
+
+    The output of `fill_value()` is computed as
+
+    ```python
+    A._fill_value().resize(A.values().shape[1:])
```
hameerabbasi, line 228: This would introduce undefined elements. Also I cannot find a `resize` method in the docs, only [`Tensor.resize_`](https://pytorch.org/docs/stable/tensors.html?highlight=resize#torch.Tensor.resize_), which would do what I said. I may be missing something though.

RFC-0004-sparse-fill-value/README.md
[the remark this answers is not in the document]
hameerabbasi: I agree with Ralf here, if it's for expert users they would understand coalescing.

RFC-0004-sparse-fill-value/README.md
hameerabbasi, line 279:
```suggestion
    A = [[1, *], [*, 4]]                  # A fill value is 2
    B = [[*, *], [7, 8]]                  # B fill value is 6
    A + B = [[1 + 6, *], [2 + 7, 4 + 8]]  # A + B fill value is 2 + 7 = 9
```

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+         tensor in COO format may contain zero values that are not
+         accounted in `nnz`
+
+       With the introduction of nonzero fill values, the misuse of
+       acronym NNZ will get worse because with nonzero fill value the
+       sparse tensor may have no zero elements, e.g. `torch.full((10,
+       10), 1.0, layout=torch.sparse_coo)` for which `nnz` would be
+       `0`.
+
+       Recommendation: stop the misuse of NNZ acronym via
```
hameerabbasi, line 393:
```suggestion
       For the reasons above, this proposal recommends renaming
```

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+
+The Scikit-Learn algorithms of the Matrix and Graph domains use sparse
+arrays from
+[scipy.sparse](https://docs.scipy.org/doc/scipy/reference/sparse.html)
+which do not record the fill value in the sparse array object
+explicitly. Instead, the corresponding algorithms make an implicit
+assumption about the fill value when processing the same SciPy sparse
+arrays.
+
+See also [the feature request of default value for scipy.sparse](https://stackoverflow.com/questions/6256206/scipy-sparse-default-value).
```
hameerabbasi, line 584: Is it possible to add `PyData/Sparse` here? 😉

hameerabbasi (commented):
I gave this a somewhat thorough review. I'll push up some English style/understanding fixes tomorrow. The English here is grammatically and structurally correct, but it's a bit hard to follow and needs some motivation.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,613 @@
...
+    A._fill_value() -> torch.tensor(1.2)
+    ```
+
+    The output of `fill_value()` is computed as
+
+    ```python
+    A._fill_value().resize(A.values().shape[1:])
+    ```
+
+    Storing the specified fill value instead of the fill value of the
```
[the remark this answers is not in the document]
pearu: Yes, it is related to Hameer's broadcasting suggestion.
Strictly speaking, [PyTorch broadcasting semantics](https://pytorch.org/docs/stable/notes/broadcasting.html) does not support broadcasting scalar values to a tensor which we need here, hence not using the term "broadcastable".

Replaced "reduced memory consumption" with "the specified fill value may have a smaller size than the fill value of the hybrid tensor", although, the other points are more important from performance point of view.

RFC-0004-sparse-fill-value/README.md
pearu: applied without "defined" to avoid tautology.

RFC-0004-sparse-fill-value/README.md
pearu: applied

RFC-0004-sparse-fill-value/README.md
pearu: ok

RFC-0004-sparse-fill-value/README.md
pearu: It applies both for hybrid as well as non-hybrid tensors.

RFC-0004-sparse-fill-value/README.md
pearu: Right, resize will produce garbage. It turns out that `torch.Tensor.resize` exists but is not documented.
The correct method to use here is `expand`. Will fix it..

RFC-0004-sparse-fill-value/README.md
hameerabbasi: In this case, perhaps change "sparse" to "hybrid" above.

RFC-0004-sparse-fill-value/README.md
pearu: This issue is now fixed with a modified example.

RFC-0004-sparse-fill-value/README.md
pearu: Fixed.

RFC-0004-sparse-fill-value/README.md
pearu: Fixed.

RFC-0004-sparse-fill-value/README.md
pearu: Applied.

RFC-0004-sparse-fill-value/README.md
pearu: I think the details why the memory contains garbage, is unimportant here.
On the other hand, the definition of undefined value does not correspond to https://en.wikipedia.org/wiki/Undefined_value , so, a better term is needed here.

pearu: Using now "uninitialized value".

RFC-0004-sparse-fill-value/README.md
pearu: Yes. Actually, the reference to PyData/Sparse replaces the stackoverflow link.

RFC-0004-sparse-fill-value/README.md
pearu: Thanks for the suggestion. Perhaps: "Extending PyTorch sparse tensors with fill value" to make it more concrete.

pearu: Or: "Adding fill value property to PyTorch sparse tensors"

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,615 @@
...
+    For example, the fill value of a (1+1)-D hybrid tensor can be
+    specified as scalar:
+
+    ```python
+    A = torch.sparse_coo_tensor(indices=[[0, 3]],
+                                values=[[.11, .12], [.31, .32]],
+                                size=(4, 2),
+                                fill_value = 1.2)
+    A.fill_value() -> torch.tensor([1.2, 1.2])
+    A._fill_value() -> torch.tensor(1.2)
```
dzhulgakov, line 220: can we just return a full-shaped tensor but restrided with strides of 0? Then I don't think you need separate `_fill_value` method.

The element-wise operators can have a fast path for the case when it's really a scalar being broadcasted to the hybrid shape. And general logic would be pretty much the same as for TensorIterator broadcasting rules.

rgommers (approved):

dzhulgakov (commented):
This is a very detailed and comprehensive proposal, thanks! Everything makes sense.

Maybe expand what exactly it'd mean for autograd? (e.g. provide a sample formula for an element-wise operation). Does it also mean that we'd need to have optimizer modification to support fill_value? (Right now there are a few optimizers that can consume sparse grads directly)

RFC-0004-sparse-fill-value/README.md
ezyang: One thing that is a little tricky here, is that if you get the restrided tensor with zero and then do an operation on it like `add_(2)`, that will cause the entire tensor to get materialized. Maybe TensorIterator shouldn't do that...

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,615 @@
...
+    A + B = [[6,  *],
+             [9, 10]]
+    # with fill value 2 + 6 = 8
+    ```
+
+10. Existing PyTorch functions that support sparse tensor as inputs,
+    need to be updated for handling the defined fill values. This will
+    be implemented in two stages:
+
+    1. All relevant functions need to check for zero fill value. If a
```
ezyang, line 291: This is trouble, if fill values live on device. If fill value lives on device, you cannot conveniently check if it is nonzero without inducing a synchronize. Indeed, the pseudocode below's use of `nonzero` would cause a sync. This is bad.

It might be possible to do some conservative analysis; for example, we can separately manage if something is "definitely zero" versus "maybe not zero" on a cpu side bit, and then have individual functions propagate it. But at that point, you might as well just upfront fix all the functions to handle fill value. Fortunately, there aren't that many; it's small enough that, for example, when I did the port from THS to ATen, I did it all in one go.

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,615 @@
...
+    2. Update the related functions to handle nonzero fill values of
+       sparse tensor inputs correctly.
+
+       For instance, consider a matrix multiplication of two sparse
+       tensors `A` and `B` with fill values `a` and `b`, respectively,
+       then the `matmul` operation can be expanded as follows:
+
+       ```python
+       matmul(A, B) = matmul(A - a + a, B - b + b)
+                    = matmul(A - a, B - b) + a * matmul(ones_like(A), B) + b * matmul(A, ones_like(B))
```
ezyang, line 316: cc @albanD doesn't this look like dual numbers? It kind of looks like dual numbers.

ezyang: Does this actually preserve sparsity? Consider this example:

```
>>> torch.tensor([[1,0,0],[0,2,0],[0,0,3]], dtype=torch.float) @ torch.ones((3, 3))
tensor([[1., 1., 1.],
        [2., 2., 2.],
        [3., 3., 3.]])
```

I'm not sure the output can be expressed in a sparse way. (I suppose the transpose can be done as one sparse dimension and one dense dimension, with a fill value of `[1, 2, 3]`)

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,615 @@
...
+    - removing the `_nnz()` method starting from PyTorch 2.0.
+
+    Alternative: Do nothing.  This is the (undocumented) approach
+    taken in Wolfram Language where [one can use "NonzeroValues" to
+    determine the number of specified
+    elements](https://community.wolfram.com/groups/-/m/t/1168496) even
+    when the fill value specified is nonzero.
+
+15. The `torch` namespace functions `arange`, `range`, `linspace`, and
+    `logspace` have `layout` argument that is not needed.
```
ezyang, line 407: Although I don't have concrete evidence for it, I don't think we should remove layout from these functions... they might come handy in the future.....

ezyang:
There isn't much in the proposal about autograd. Although I don't think fill values necessarily make the situation any worse than they were before, I think they do deserve some discussion. In particular, suppose I do a function f(x) where x has a non-zero fill value. What should the fill value of the computed `grad_x` be?

I think the answer is simply that the gradient should be derived off of the computations that involved the fill value directly, and the sparsity pattern should be preserved. But it would be good to see this worked out in more detail, and to verify that desirable mathematical properties still hold.

RFC-0004-sparse-fill-value/README.md
pearu: No, in general, the matmul of sparse tensors with nonzero fill values will result in a dense tensor (read: a tensor in sparse format but with a small number of equal elements).

While this is going to be slightly off-topic, but when there exist applications that require such a matmul, I can think of one way to preserve the sparsity from matmul: a lazy evaluation. It would be like uncoalesced tensor but at the level of tensors, not tensor elements.

For instance, the result of `A @ B` would be an object, say, LazyTensor that holds three sparse tensors: `(A-a) @ (B-b)`, `ones_like(A) @ (a*B)`, and `(b*A) @ ones_like(B)`. Each of these tensors can be stored memory-efficiently as sparse tensors while adding these would lead to a dense tensor (but we are going to postpone the addition). So, when one has
```
A @ B -> C = LazyTensor((A-a) @ (B-b), ones_like(A) @ (a*B), (b*A) @ ones_like(B))
```
then computing a matrix-vector product `C @ v` boils down to computing the sparse matrix-vector product with each of the three sparse tensors and `v`:
```
LazyTensor(C1, C2, C3) @ v -> C1 @ v + C2 @ v + C3 @ v
```
so that the postponed addition is realized with the results of matrix-vector products.

ezyang: OK. We're generally amenable to laziness (see also https://github.com/pytorch/pytorch/issues/43563 ) and it is easier to do this in sparse as we have no concept of views, which make laziness harder to do. But out of scope for this RFC :)

mruberry:
Hey @pearu, thanks for this write-up!

I appreciate your thinking on sparse tensors and adding a "fill value" to them seems like a mathematically clear and consistent way of defining them. However, I'm not sure if an implementation will actually give users clear and consistent behavior. That is, will users understand the difference between sparse tensors with and without fill values, how to preserve sparsity, and which operations are/aren't performant or memory-saving when setting a fill value? Let me try to detail that query with some sub-questions:

- How do we define the behavior of sparse tensors without a fill value? These are not consistently interpreted as having a value of zero today.
- How do we think about the gradient of sparse tensors without a fill value? What if the gradient specifies more values than the sparse tensor?
- How do we think about the gradient of sparse tensors with a fill value? What if the gradient specifies more values than the sparse tensor?
- If I have sparse A and strided B, what does A * B return? I would be concerned if A's having or not having a fill value could change properties of the returned tensor. How intuitive would that be? 

I understand the desire to implement a fill value for some cases, but I'd like to get a better sense for the mathematics behind the no fill value definition of a sparse tensor, and I wonder if we shouldn't apply more guardrails to sparse tensors with fill values to protect users from getting into trouble. Also, as mentioned, I'd like to get a better understanding for how we treat sparse tensors without a fill value. Should we consider them masked tensors and let functions have their own interpretations of that mask?

rgommers:
> Also, as mentioned, I'd like to get a better understanding for how we treat sparse tensors without a fill value. Should we consider them masked tensors and let functions have their own interpretations of that mask?

It may be good to answer this by prioritizing a good rewrite of the docs (https://github.com/pytorch/pytorch/issues/44635) that answers this, rather than go into detail on this PR. That has to be done anyway (for 1.7.0), and doing it first will make this RFC easier to fit into the picture.

ezyang:
Another thought that @mruberry and I had, was that although fill values are a more conservative extension in a semantic sense, as they maintain the correspondence between sparse and dense tensors, going straight to masked tensors, and adding fill values as an extension on top of it, might be a path that reduces the amount of absolute work we have to do (and in particular, may help us avoid having to add a bunch of complicated logic for handling fills in scenarios where no one really cares.)

ngimel:
I agree with Ed. This RFC contains one valid example of when non-zero fill value is helpful (random sequence of rare events with nonzero mean), others (exp, log, softmax etc), at least in the neural network context, are better handled by masking semantics. Masking semantics also provides necessary restrictions on the sparsity pattern of the gradient, so the answer to @mruberry's question "What if the gradient specifies more values than the sparse tensor" is "It can't".

RFC-0004-sparse-fill-value/README.md
```diff
@@ -0,0 +1,615 @@
...
+    default fill value of a hybrid tensor, see points 5 and 6 below.
+
+3.  PyTorch functions that have the `layout` argument may use a different
+    `fill_value` when constructing a sparse tensor as defined by the
+    following table:
+
+    | Function              | `fill_value` of returned sparse tensor |
+    | :-------------------- | :------------------------------------- |
+    | `torch.empty`         | uninitialized value                    |
+    | `torch.empty_like`    | uninitialized value                    |
```
pearu, line 118: Currently, `torch.empty(layout=torch.sparse_coo)` is equivalent to `torch.zeros(layout=torch.sparse_coo)`.
For backward compatibility, the default fill value for empty ought to be 0.

albanD: For sparse Tensor we actually zero-out the memory in `empty()`? Because for dense Tensors this is not true.

pearu: The `empty` equivalence applies only when using sparse layout, it does not hold for strided tensors indeed.
In the case of sparse tensors, there is nothing to zero-out in `zeros`: sparse `zeros` is sparse `empty` by the definition (the default fill value is `0`).

pearu:
Just for heads up, I'll respond to review comments shortly. Sorry for the delay!

RFC-0004-sparse-fill-value/README.md
pearu: @ezyang Having the fast "definitely zero" test would be advantageous even after fixing all the functions to handle fill value. This is because most Linear Algebra functions would always do the zero-test in order to use the fastest and most frequently used zero-fill-value-pathway.

An idea: let the fill value always live in the CPU memory so that the zero-test will be fast. When an algorithm needs to use fill value from a device, the algorithm makes the to-device copy.

To make coding easier, one could use unified CUDA memory only for the fill value when the sparse tensor `values` uses CUDA device. Although, other non-CUDA devices would not gain from this and explicit copy-fill-value-to-device would be still needed.

RFC-0004-sparse-fill-value/README.md
pearu: Labelling this proposal as rejected.

cpuhrsch:
I'm still working my way into this topic, so please have some patience as I catch up on a many year-long effort.

When we say "masking semantics" is this maybe too implementation specific?

To some extend a "mask" could perhaps also be thought of as introducing nullability. The mask can be used to determine which value is and is not null, but it's not the only metadata representation that can achieve this (even though it might be convenient for many kernels).

Further, is a mask truly necessary for the use cases that require this kind of nullability? It can introduces a lot of branching, which I could see particularly difficult to support on GPUs in a generic sense. 

Also, in some cases a user might want different mask semantics. Let's take addition as an example.

a) fail when the masks don't agree: addition of two tensors encoding variable length sentences
b) take the intersection of masked values (e.g. MHA input mask + attention mask):  replace masked-out values with the identity
c) take the union of masked values: if one of the operands values is maked out so will be the corresponding results. This is the masked array's behavior in numpy, but I don't know a use case for this.

cpuhrsch:
Sure, but the semantics of dense tensor + mask will still be defined per operator. I think it's actually an implementation independent generalization (e.g. we might still want to use sparse storage layouts even when the object is semantically a dense tensor + mask). As such we should be clear on what exactly we actually want to achieve with these dense tensor + mask semantics and what makes them uniquely useful.

PR #9, opened 2020-10-13
Structured kernel definitions RFC

ezyang:
Signed-off-by: Edward Z. Yang <ezyang@fb.com>

### DIFF
--- a/RFC-0005-structured-kernel-definitions.md
+++ b/RFC-0005-structured-kernel-definitions.md
@@ -1,0 +1,323 @@
+# Structured kernel definitions in PyTorch
+
+This is a proposal for a new code generation facility for writing
+kernels in PyTorch, where we will automatically generate
+easy-to-get-wrong boilerplate for functional (`add`), inplace (`add_`) and
+out (`add_out`) variants of functions, as well as common code (device
+guards, version counter tracking). The net result is you only need to
+write a shape checking function and an out-kernel when writing a
+function.
+
+**NOW HAS A PROTOTYPE** https://github.com/pytorch/pytorch/pull/45277 
+
+## Goals
+
+* **Be an opt-in mechanism**: always OK to continue to write code as it
+  is today. This ensures that third-party backend extenders also can
+  make things work without making use of codegen
+
+* **Reduce shape checking boilerplate**: make it easier to share common
+  shape checking code between CPU and CUDA implementations, as well as
+  with out-of-tree backend implementations
+
+* **Reduce functional/inplace/out boilerplate**: avoid having to write
+  foo, foo_, foo_out variants for every function; make it harder to
+  forget to add foo_out variant when it is appropriate
+
+* **Be able to run forward shape computation without kernel**: so
+  compilers and static runtimes can easily compute sizes of all elements
+  in the framework without having to actually run the computation
+
+* **Provide entry point for static runtime**: provide public API for
+  accessing operators directly, bypassing output allocation, shape
+  checking, device guards.
+
+* **Bypass dispatcher overhead**: in high performance cases, e.g., core
+  framework operators, reduce the number of redispatches to improve
+  performance
+
+* **Unify with version counter bumps and view metadata tracking**: these
+  are currently done in autograd generated code but must be performed
+  unconditionally even when autograd is disabled. This logic to be
+  incorporated with logic here
+
+* **Work with mobile selective build**: mobile selective build is
+  implemented in codegen, and implementation strategy must be compatible
+  with design constraints in mobile
+
+## Non-goals
+
+* **Directly handle TensorIterator**: TensorIterator is a uniquely
+  complicated mechanism, and we think it should be designed for on its
+  own.  However, we think that this is an important case and should
+  be handled sooner rather than later.
+
+* **Be zero runtime cost all the time**: We are willing to give up some
+  runtime efficiency for cleaner code. We are willing to give up
+  efficiency *by default* for out of tree implementers, unless they opt
+  in to higher performance. There is always an escape hatch to be high
+  performance if absolutely necessary
+
+* **No codegen**: As long as it is possible to implement things out of
+  tree (at the cost of human understandable boilerplate), codegen is a
+  reasonable strategy for implementing features we need
+
+* **Major format change to native_functions.yaml**: We could do this. We
+  are choosing not to in order to reduce the degrees of freedom in what
+  changes we make to `native_functions.yaml`.
+
+## Summary
+
+* Core
+
+    * Define a new API for shape checking functions. Shape checking
+      functions are called from generated code that implements
+      functional/inplace/out variants of functions.
+
+    * Define a new API for kernels. Static kernel API does not do shape
+      checking, output is always preallocated at correct size. This API
+      is public and a suitable entrypoint for static runtime. Static
+      kernels are called from generated code like above.
+
+    * Code generate meta functions from shape checking functions to
+      provide public API for running shape computations without any
+      kernel.
+
+    * Generated code is augmented to do version counter bumps and view
+      tracking
+
+* Extensions
+
+    * Add a new dispatch key (name tbd) which contains shape checking,
+      device guard, version counter bump. External backend
+      implementations of PyTorch operators which do not explicitly opt
+      out of this dispatch key will get this logic applied to them. This
+      key is skipped for core operators, as this checking logic is fused
+      directly into the backend kernel. External backends can also opt
+      to fuse this logic in.
+
+## Details
+
+Let’s suppose you want to write a new operator from scratch. The first
+thing you will do is add a native_functions.yaml declaration for it.
+Let’s take `upsample_nearest1d` as the example for this post.  You will
+write an entry like this, describing the *functional* version of this
+operator:
+
+```
+- func: upsample_nearest1d(Tensor self, int[1] output_size, float? scales=None) -> Tensor
+  dispatch:
+    CPU: upsample_nearest1d_cpu
+    CUDA: upsample_nearest1d_cuda
+```
+
+Because you are a conscientious implementor, you also want to provide an
+out= variant of this function. This version gets a separate entry:
+
+```
+- func: upsample_nearest1d.out(Tensor self, int[1] output_size, float? scales=None, *, Tensor(a!) out) -> Tensor(a!)
+  dispatch:
+    CPU: upsample_nearest1d_out_cpu
+    CUDA: upsample_nearest1d_out_cuda
+```
+
+Ordinarily, these two declarations would cause function signatures for
+`upsample_nearest1d_cpu`, etc., to be generated into NativeFunctions.h,
+and then you would go ahead and implement them.  However, we propose a
+new *structured* format for writing kernels. We’ll do this by marking
+the out version of this operator as structured and deleting dispatch
+entries from the functional version (the functional operator is
+*implicitly* associated with the out-of-place version in the same way
+derivatives.yaml schemas are matched with their functional/inplace/out
+variants):
+
+```
+- func: upsample_nearest1d(Tensor self, int[1] output_size, float? scales=None) -> Tensor
+  # [NEW] dispatches for this function are omitted
+
+- func: upsample_nearest1d.out(Tensor self, int[1] output_size, float? scales=None, *, Tensor(a!) out) -> Tensor(a!)
+  structured: True  # [NEW]
+  dispatch:
+    CPU: upsample_nearest1d_structured_cpu
+    CUDA: upsample_nearest1d_structured_cuda
+```
+
+Structured definitions require a different set of functions to be
+written to implement the operator.  In this particular case, the
+functions you have to implement are:
+
+```
+namespace meta {
+ TensorMeta upsample_nearest1d(
+ const Tensor& self, IntArrayRef output_size, optional<double> scales);
+ }
+ 
+ namespace native {
+ // Precondition: out is an allocated and appropriately sized tensor;
+ // all shape checks have passed, device guards have been set,
+// version counter bumps are all handled, etc...
+ void upsample_nearest1d_structured_cpu(
+const TensorMeta& out_meta, const Tensor& out, const Tensor& self, IntArrayRef output_size, optional<double> scales);
+ void upsample_nearest1d_structured_cuda(
+const TensorMeta& out_meta,  const Tensor& out, const Tensor& self, IntArrayRef output_size, optional<double> scales);
+ }
+```
+
+The code generator then generates the boilerplate code to put these
+functions together:
+
+```
+namespace native {
+
+Tensor upsample_nearest1d_cuda(const Tensor& self, IntArrayRef output_size, optional<double> scales) {
+  CUDADeviceGuard g(self.device());
+  TensorMeta expected = meta::upsample_nearest1d(self, output_size, scales);
+  Tensor result = native::empty(expected.sizes(), expected.options());
+  upsample_nearest1d_structured_cuda(expected, result, self, output_size, scales);
+  return result;
+}
+
+Tensor& upsample_nearest1d_out_cuda(Tensor& result, const Tensor& self, IntArrayRef output_size, optional<double> scales) {
+  // In event of multiple tensor arguments, code generation should
+  // be responsible for making sure all devices are consistent
+  CUDADeviceGuard g(self.device());
+  TensorMeta expected = meta::upsample_nearest1d(self, output_size, scales);
+  TORCH_CHECK(expected.options == result.options());
+  result.resize_(expected.size());
+  upsample_nearest1d_structured_cuda(expected, result, self, output_size, scales);
+  increment_version(result);
+  return result;
+}
+
+// CPU follows similarly
+
+}
+
+```
+
+Notice that the generated boilerplate here handles several concerns:
+
+* It properly sets up device guards for the kernel
+
+* For the `_out` variant, it tests that the inferred tensor options are
+  consistent with the passed in tensor
+
+* It centralizes the PyTorch convention that improperly sized out=
+  arguments get resized to fit the output
+
+* It handles version counter bumping for mutations
+
+* It is carefully written so that there are no extra redispatches; all
+  calls are direct non-dispatched function calls
+
+It’s also worth noting what things the boilerplate here does not handle:
+
+* It forces the allocation of a TensorMeta struct per output tensor,
+  before the final allocation of the output tensor. You could save a few
+  instructions by eliminating this struct entirely. However, I could not
+  think of an elegant API for doing so.
+
+In absolute terms, the boilerplate code saved here is not all that
+large; only several lines. However, when multiplied over the number of
+kernels in PyTorch, and the subtlety of remembering to handle all of
+these issues, and we think the use of code generation to automatically
+generate this boilerplate is worth it.
+
+The functionality code generated here is not made available to other
+backends. This is problematic, so we also generate a separate dispatch
+key handler which will automatically handle shape checking and other
+concerns for backends that are not concerned with performance:
+
+```
+// Name TBD; for this example we will call it DispatchKey::Common, as the
+// functionality here is common to all backends. This is an alias key
+// that resolves CommonXLA/CommonMSNPU/... in the same way as Autograd.
+
+Tensor upsample_nearest1d_common(const Tensor& self, IntArrayRef output_size, optional<double> scales) {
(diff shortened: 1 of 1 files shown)


RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+ }
+ 
+ namespace native {
+ // Precondition: out is an allocated and appropriately sized tensor;
+ // all shape checks have passed, device guards have been set,
+// version counter bumps are all handled, etc...
+ void upsample_nearest1d_structured_cpu(
+const TensorMeta& out_meta, const Tensor& out, const Tensor& self, IntArrayRef output_size, optional<double> scales);
+ void upsample_nearest1d_structured_cuda(
+const TensorMeta& out_meta,  const Tensor& out, const Tensor& self, IntArrayRef output_size, optional<double> scales);
```
bhosmer, line 162: maybe worth noting somewhere in here that you're limiting the example to CPU and CUDA for expository purposes? the use of explicit dispatch keys above should probably imply to the careful reader that this setup is parameterized over all backend dispatch keys in the usual way (IIUC), but if so probably useful to make it explicit

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+```
+
+## Discussion
+
+* Why not just use the Common dispatch key for everything?
+
+    * Performance reasons. Introducing the common key would induce an
+      extra redispatch, which at time of writing would give up quite a
+      bit of performance due to dispatch overhead, for no particularly
+      good reason.
```
bhosmer, line 286: I think it's worth thinking carefully about whether this is the best way to make structured kernels available out of tree, given the modality it introduces, risk of drift, perf handicap (albeit diminishing over time) it saddles out of tree backends with, etc.

E.g. once the rest of the codegen has been ported, it's not absurd to imagine fitting it with a frontend that can take inputs other than the in-tree yaml...

ezyang: The common dispatch key is primarily oriented towards backend implementers.

Let's suppose for a moment, that an overriding design goal is safety/correctness first, with the ability to opt into performance.  Then it feels feel like I am forced to introduce the Common dispatch key, because without it, an average backend overrider has to faithfully replicate all of the functionality that we have otherwise fused into CPU/CUDA operators (shape checking, device guards, and with some refactors coming soon, it will also include version counter bumps). It's unrealistic to expect a backend implementer to actually manage all of this without code generation.

That being said, there is a certain optionality to the Common dispatch key. We don't have to implement it (and indeed, in the current posted PR, it is not implemented), and if it is not implemented, the burden is simply on backend implementers to implement all of the necessary scaffolding (which is the de facto situation today). If, for example, we pivoted to publishing code generation for backend implementers, that would alleviate or perhaps eliminate the need for a common dispatch key.

The common dispatch key is mostly irrelevant for custom operators, since average use of common operators is via catch all registration and there will not be any common dispatch key in any case.

bhosmer: Yeah, agree on all the motivations. My pitch would be to (wait and) audition a lightly parameterized codegen pipeline as an alternative, once the rewrite is complete.

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+    * Define a new API for kernels. Static kernel API does not do shape
+      checking, output is always preallocated at correct size. This API
+      is public and a suitable entrypoint for static runtime. Static
+      kernels are called from generated code like above.
+
+    * Code generate meta functions from shape checking functions to
+      provide public API for running shape computations without any
+      kernel.
+
+    * Generated code is augmented to do version counter bumps and view
```
smessmer, line 86: This means we are using this mechanism for all ops, even ops that don't have out or inplace variants, right? Otherwise those ops wouldn't get device guards or version counter bumps.

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+  own.  However, we think that this is an important case and should
+  be handled sooner rather than later.
+
+* **Be zero runtime cost all the time**: We are willing to give up some
+  runtime efficiency for cleaner code. We are willing to give up
+  efficiency *by default* for out of tree implementers, unless they opt
+  in to higher performance. There is always an escape hatch to be high
+  performance if absolutely necessary
+
+* **No codegen**: As long as it is possible to implement things out of
```
smessmer, line 61: As long as we use this mechanism internally for all ops, we have meta functions for them defined and backends adding new kernels don't need to care about it. Only backends that add new operators would have to use it. Did I understand this correctly?

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+    * Code generate meta functions from shape checking functions to
+      provide public API for running shape computations without any
+      kernel.
+
+    * Generated code is augmented to do version counter bumps and view
+      tracking
+
+* Extensions
+
+    * Add a new dispatch key (name tbd) which contains shape checking,
```
smessmer, line 91: This dual approach makes the system a bit complex, it can be hard to figure out where version counter bumps, device guards or shape checking happen for a corresponding op, especially when you're not familiar with this system (yet).

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+    CUDA: upsample_nearest1d_out_cuda
+```
+
+Ordinarily, these two declarations would cause function signatures for
+`upsample_nearest1d_cpu`, etc., to be generated into NativeFunctions.h,
+and then you would go ahead and implement them.  However, we propose a
+new *structured* format for writing kernels. We’ll do this by marking
+the out version of this operator as structured and deleting dispatch
+entries from the functional version (the functional operator is
+*implicitly* associated with the out-of-place version in the same way
```
smessmer, line 130: This implicit connection makes reasoning about entries in `native_functions.yaml` harder since they're not independent entries anymore.

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+derivatives.yaml schemas are matched with their functional/inplace/out
+variants):
+
+```
+- func: upsample_nearest1d(Tensor self, int[1] output_size, float? scales=None) -> Tensor
+  # [NEW] dispatches for this function are omitted
```
smessmer, line 136: hm maybe we should make this explicit by also adding `structured: True` here, to make the information more local and easier to read. Otherwise you have to look at other entries in native_functions.yaml to know what this entry is actually doing. Omitting dispatch keys is already a valid kernel definition even if it's not structured, so by just looking at this one it would be ambigious.

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+ }
+```
+
+The code generator then generates the boilerplate code to put these
+functions together:
+
+```
+namespace native {
+
+Tensor upsample_nearest1d_cuda(const Tensor& self, IntArrayRef output_size, optional<double> scales) {
```
smessmer, line 172: these should be easy to generate with templates instead and it should be possible to do that in a readable way

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+backends. This is problematic, so we also generate a separate dispatch
+key handler which will automatically handle shape checking and other
+concerns for backends that are not concerned with performance:
+
+```
+// Name TBD; for this example we will call it DispatchKey::Common, as the
+// functionality here is common to all backends. This is an alias key
+// that resolves CommonXLA/CommonMSNPU/... in the same way as Autograd.
+
+Tensor upsample_nearest1d_common(const Tensor& self, IntArrayRef output_size, optional<double> scales) {
```
smessmer, line 236: We could offer a simple metaprogram for them so they don't have to manually write this for each op. But if I understand correctly, they only have to write this for ops they introduce right? Ops from native_functions.yaml that the backend only extends will already have shape checking etc. through our codegen right?

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,323 @@
...
+      generated here is already duplicated at the source level today
+      (for example, many operators have shape checks factored into
+      header files that are then shared between two operators).
+
+* I don’t like codegen.
+
+    * An earlier version of this proposal had the boilerplate
+      generated using C++ templates rather than codegen. However, we
+      think the formulation in this proposal is superior under the
+      constraint that mobile selective build must keep working, as we
```
smessmer, line 300: I don't understand the connection to mobile selective build yet

smessmer (commented):
I think some of the weirdness like making native_functions.yaml entries depend on each other or having two ways of doing things (dispatch key and fusing into kernels) is because you designed with the constraint of not changing the fact that the three operator overloads (regular, inplace, out) are actually registered as separate operators to the dispatcher.

Have we considered an alternative solution where the dispatcher only knows about the out overload and the regular and inplace variants are generated in the frontend before the call to the dispatcher? That would also make it extensible for all ops and take the burden away from backend implementors.

RFC-0005-structured-kernel-definitions.md
ezyang: Yeah, this is an orthogonal change that we can do with or without structured kernels.

RFC-0005-structured-kernel-definitions.md
ezyang: That's right. So if you don't care about performance, you don't even have to lift a finger.

RFC-0005-structured-kernel-definitions.md
ezyang: The current expressivity problems:
1. Guard logic is weird and special-casey in codegen right now. We probably *should* be able to simplify it a bit, but right now it would be quite difficult to faithfully replicate the logic in a template
2. The upcoming version counter bumps will be difficult to do without out-of-band information about what arguments are mutable or not (right now you can check this using `Tensor&` but when we fix everything to uniformly be `const Tensor&` you'll lose this type info)

RFC-0005-structured-kernel-definitions.md
ezyang: Yes

RFC-0005-structured-kernel-definitions.md
ezyang: We can talk about this in the meeting. It's constraint solving from the problem "mobile requires registrations to be a separate compilation unit from kernels"

RFC-0005-structured-kernel-definitions.md
smessmer: wait I think I misunderstood this. To get device guards, you'll still have to go through the dispatch key for those backends, but not if calling into internal backends like CPU. How do you selectively make that dispatch key fallthrough for some backends but not for others?

ezyang: I claimed this would work in our meeting, but rereading your comment, I take it back, you really do need a per backend key here :(

salexspb:
This is very exciting! Looking forward having this structure. 
Ability to infer shapes statically and pre-allocate all the memory to be passed on later to _out kernel versions is going to bring PyTorch runtime to the next level,  I think. 

Do you guys plan to support all GPU standard kernels using this new framework? (so they have shape inference and _out versions). 
And what about custom ops / classes, do you plan to support them in the kernel definition framework as well? Ideally I would like to be able to have 100% static models via using a mix of built-in _out ops versions and custom ops (also accepting external memory and providing shape checking / inference functionality) one may implement.

ezyang:
> Do you guys plan to support all GPU standard kernels using this new framework? (so they have shape inference and _out versions).

Yeah. The aspiration is every kernel in PyTorch is in this framework. That's gonna be a lot of work, but hopefully we can lay the groundwork and then roll it out as we go.

> And what about custom ops / classes, do you plan to support them in the kernel definition framework as well?

This is not entirely settled yet, but the intention for the section at https://github.com/pytorch/rfcs/pull/9#discussion_r512447736 was to make this possible. So if you are doing a custom op, you now define (e.g.) three parts: a CPU part, a CUDA part, and the static shape checking part, and the framework would put it all together for you.

One extra thing that I'd add, though, is when I've been chatting with other static runtime people at FB, they seem to want strange things like being able to run the static inference really fast at the beginning of each run to work out the preallocation. At least the first iteration, the API for actually running these shape computations won't be particular fast, and will be mostly useful for offline use cases.

RFC-0005-structured-kernel-definitions.md
```diff
@@ -353,22 +445,88 @@ TORCH_LIBRARY_IMPL(aten, Meta, m) {
...
+  dispatch:
+    CPU: upsample_nearest1d_structured_cpu
+    CUDA: upsample_nearest1d_structured_cuda
+```
+
+Now you can simply construct it appropriately in your function
+definitions:
+
+```
+TORCH_META_FN(add) (
```
bhosmer, line 518: `TORCH_META_FUNC` I think

RFC-0005-structured-kernel-definitions.md
```diff
@@ -146,127 +145,208 @@ matched with their functional/inplace/out variants):
...
 namespace native {
 
+struct upsample_nearest1d_cuda : public meta::upsample_nearest1d {
+  void impl(const Tensor& self, IntArrayRef output_size, optional<double> scales); // user defined
+};
+
+// functional implementation
+
+// NB: set_output could be devirtualized with CRTP, but for now we don't do this
+struct upsample_nearest1d_cuda_functional final : public upsample_nearest1d_cuda {
```
bhosmer, line 229: Missing constructor?

ezyang: It turns out I can't have meta functions be constructors, because they want to call virtual methods but virtual method calls are not valid in constructors. So I am moving theme to be their own method.

ezyang:
Some new developments from putting the rubber to the road, implementation-wise:

* Need a way to mark some dispatch keys as "legacy" and don't opt them in to structured kernels. This is very salient for functions like add which have both dense and sparse implementations; sparse is not supported by structured (yet, anyway), so need to opt it out of structured while having everything else be structured
* There are a lot of direct calls to `native::` functions; this API point no longer is available when something is migrated to structured. Need to somehow restore this API point. Note that previously combined CPU/CUDA paths will be separated in structured universe, so this is going to be a BC break.

vadimkantorov:
will this help in porting Caffe2 kernels into PyTorch? I think there ops are more "structured" (but maybe not in the way of this proposal?)

ezyang:
@vadimkantorov Unfortunately I don't think it will help all that much :( Or at the very least, I wasn't designing for this.

ezyang:
> The *this is deeply weird until you know the context this code is being defined in... it makes me think we might want the macro names to avoid overselling that it's a freestanding function. It wouldn't need to literally say "you're looking at a method" but like TORCH_STRUCTURED_IMPL or something.

Fair enough. We can use your naming!

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,644 @@
...
+rather than a `const Tensor&`.  This convention has lead to no end
+of confusion for kernel writers, who incorrectly surmise that given
+a mutable reference `Tensor& out`, one may assign the output by
+writing `out = ... some expression ...`  (this doesn't work).
+
+The absence of the const modifier is currently relied upon by
+template metaprogramming machinery (to detect if arguments are out
+tensors or not); however, because the implementations of structured
+kernels are a layer below the operator registration layer, the
+const modifier can be eliminated from the `TORCH_IMPL_FUNC` API
```
bhosmer, line 566: I think you mean the const modifier can be added, not eliminated, right? Like, out arguments can now be typed as `const Tensor&` in `TORCH_IMPL_FUNC` declarations, as in
```
TORCH_IMPL_FUNC(upsample_nearest1d_out_cpu) (
    const Tensor& input,
    IntArrayRef output_size,
    c10::optional<double> scales,
    const Tensor& output             // currently just Tensor& output
) {
  upsample_nearest1d_kernel(kCPU, output, input, scales);
}
```
Aside: it happens that `upsample_nearest1d_kernel` itself takes `Tensor& output`, which might hint at additional porting work for at least some kernels to adopt this change.

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,644 @@
...
+that CPUTensors can be created from `const Tensor&` without incurring
+a reference count bump).
+
+One question is wheter or not the existence of CPUTensor means we should
+eliminate the `at::cpu::` namespace (as they serve near equivalent purposes;
+if you have functions which support CPUTensor, simply (unsafely) cast
+your Tensor to a CPUTensor and then utilize the regular API.)  One
+possible argument for retaining the `at::cpu::` namespace is that these
+functions are guaranteed to bypass dispatching, whereas other functions
+may implicitly downcast to `Tensor` and do an optimized call.
```
bhosmer, line 606: upcast

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,644 @@
...
+The primary consequence of making this change immediately is we must
+immediately create a CPUTensor class with enough methods to cover the
+usual surface area (even if those methods don't apply any performance
+optimization).  With code generation this should not be too much code.
+This would also require the creation of a CPUTensorRef class to ensure
+that CPUTensors can be created from `const Tensor&` without incurring
+a reference count bump).
+
+One question is wheter or not the existence of CPUTensor means we should
+eliminate the `at::cpu::` namespace (as they serve near equivalent purposes;
```
bhosmer, line 601: `at::cpu` seems much less risky around the edges than `CPUTensor`[1] - if they serve near equivalent purposes, what's an example that motivates adding CPUTensor either instead of or in addition to `at::cpu`?

[1] stemming from CPUTensor not reeeally being a subtype of Tensor. A stray call to something like `set_storage_and_dtype` would damage a CPUTensor that's been upcast to a Tensor (or provoke a runtime error, if we've put safety measures in place).

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,644 @@
...
+* View operations
+* Factory functions
+* `copy_`
+
+Since non-structured kernels simply operate at a lower level of
+abstraction, in principle, it is not a big deal if some operators
+never become structured; to make an analogy, sometimes you have to
+write assembly, and as long as it is not too frequent, there is not
+too much to be gained from trying to extend the functionality of your
+system to expunge these entirely.
```
bhosmer, line 624: A counterargument to consider here I think is that making structured kernels ubiquitous would allow other system capabilities to be defined in terms of them - e.g., shape analysis, memory planning.

The answer wouldn't be to torture everything into having in/func/out variants, but probably to add a handful of other little composite kernel system variations, each with its own codegen and triggering annotation in native_functions.yaml. Factory functions are an obvious fit for this kind of approach, but maybe they're the only good fit? Haven't thought it through any further.

RFC-0005-structured-kernel-definitions.md
ezyang: It is an ergonomics thing (but important ergonomics): with `at::cpu`, you have to remember to use the CPU only function (and you have to make sure not to accidentally call it with a non-CPU tensor). CPUTensor ensures that the CPU tensor is used, and you only need to prove you have CPU tensor when you initially construct the CPUTensor.

RFC-0005-structured-kernel-definitions.md
ezyang: Well, this certainly *would* work.  What is less certain is if the number of operators that would be covered by the system variation warrant the variations. Well, at least we can delay designing this for later...

RFC-0005-structured-kernel-definitions.md
```diff
@@ -0,0 +1,720 @@
...
+
+namespace native {
+  // Precondition: out is an allocated and appropriately sized tensor;
+  // all shape checks have passed, device guards have been set,
+  // version counter bumps are all handled, etc...
+  /* macro expands to: void upsample_nearest1d_structured_cpu::impl( */
+  TORCH_IMPL_FUNC(upsample_nearest1d_structured_cpu) (
+    const Tensor& self, IntArrayRef output_size, optional<double> scales, const Tensor& out
+  );
+  /* macro expands to: void upsample_nearest1d_structured_cpu::impl( */
```
byronyi, line 173: `s/upsample_nearest1d_structured_cpu/upsample_nearest1d_structured_cuda/`?

PR #10, opened 2020-11-26
RFC-0006: a PyTorch conda distribution

rgommers:
The main aims of this proposal are:

1. to make it easier for package authors to release packages that depend on PyTorch
2. to make it easier for conda users to install a set of packages that depend on PyTorch
3. ensure this set of packages is integration tested

Credits for the initial idea of this "integration tested distribution" go to @jph00. Thanks also to @pearu and @hameerabbasi for feedback on an earlier version, on the conda-forge section in particular.

More details on how the CI for this should work can be added, but that can be done after there's agreement that this proposal is the right direction to go in.

_EDIT: up-to-date rendered version at https://github.com/rgommers/rfcs/blob/conda-distribution/RFC-0006-conda-distribution.md_

### DIFF
--- a/RFC-0006-conda-distribution.md
+++ b/RFC-0006-conda-distribution.md
@@ -3,10 +3,9 @@
 |            |                 |
 | ---------- | --------------- |
 | Authors    | Ralf Gommers    |
-| Status     | Draft           |
+| Status     | Rejected        |
 | Type       | Process         |
 | Created    | 2020-11-26      |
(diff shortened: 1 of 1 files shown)


rgommers (commented):
@malfet and I discussed this proposal. We agreed that, given the progress conda-forge is making and GPU hardware for use in CI being added in the near future, it's better to help out with that effort.

So I'll merge this PR with status Rejected.

PR #11, opened 2020-12-11
RFC-0007: Forward AD

albanD:
This rfc is capturing the discussions that happen in the main issue https://github.com/pytorch/pytorch/issues/10223 as well as private conversations with different people.

It presents the design idea and implementation plan for forward AD withing PyTorch.


ezyang:
Add a rendered link! https://github.com/albanD/rfcs/blob/rfc-0007/RFC-0007-forward-AD.md

PR #16, opened 2021-03-03
RFC-0010: TensorRef

ezyang:
This proposal introduces a new class `TensorRef`, to replace all
places in our codebase where we currently use `const Tensor&`.  The
distinguishing characteristics of this class are:

1. It is non-owning
2. It is as safe as other by-value reference types (like `c10::ArrayRef` or
   `std::string_view`)
3. It is implicitly convertible (with some exceptions) to `const Tensor&`
   (i.e., it can be introduced incrementally)

[Rendered](https://github.com/pytorch/rfcs/blob/tensor-ref/RFC-0010-tensor-ref.md)

Signed-off-by: Edward Z. Yang <ezyang@fb.com>

### DIFF
--- a/RFC-0010-tensor-ref.md
+++ b/RFC-0010-tensor-ref.md
@@ -1,0 +1,111 @@
+# TensorRef abstraction in C++
+
+This proposal introduces a new class `TensorRef`, to replace all
+places in our codebase where we currently use `const Tensor&`.  The
+distinguishing characteristics of this class are:
+
+1. It is non-owning
+2. It is as safe as other by-value reference types (like `c10::ArrayRef` or
+   `std::string_view`)
+3. It is implicitly convertible (with some exceptions) to `const Tensor&`
+   (i.e., it can be introduced incrementally)
+
+## Motivation
+
+* In order to call a function that accepts `const Tensor&`, you must
+  somehow have a `Tensor` at hand.  However, `Tensor` is always an
+  owning value type, and sometimes you only have a non-owning reference
+  to a Tensor, but not a reference to Tensor itself (e.g.,
+  `TensorImpl*`).  In these cases, you cannot conveniently call the
+  function without first inducing a refcount bump (to establish an
+  owning `Tensor` object, to pass by reference).
+
+* For performance reasons, IValue `toTensor` returns a `const Tensor&`.
+  For this to be possible, IValue must internally store a `Tensor`
+  object (otherwise, there is nothing to point the reference to.)  This
+  induces a dependency from IValue header to Tensor header, which can
+  cause problems in some situations (as Tensor class has a large API and
+  thus has many dependencies on many other classes.)
+
+* `const Tensor&` implies a double indirection, as `Tensor` is itself
+  already a pointer.  A TensorRef type implemented internally as a
+  `TensorImpl*` would eliminate this double indirection.  (NB: The
+  proposal below doesn't actually solve this problem, due to the definition
+  of Itanium ABI.)
+
+## Proposal
+
+Add the following class to PyTorch:
+
+```
+struct TensorRef {
+    Tensor base_;
+    TensorRef(TensorImpl* impl) : base_(TensorPtr::reclaim(impl)) {}
+    ~TensorRef() {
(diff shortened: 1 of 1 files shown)


RFC-0010-tensor-ref.md
```diff
@@ -0,0 +1,119 @@
...
+};
+```
+
+Main points:
+
+* This class is implemented as containing an actual Tensor object, but
+  we unsafely use reclaim/release to ensure that no refcount bump occurs
+  on construction/destruction of the object.  We verified with Godbolt
+  that the compiler is able to eliminate the `base_` destructor.  This
+  makes it possible to take out a `const Tensor&`, which makes it easier
```
smessmer, line 62: as long as the `Tensor` is purely internal to `TensorRef`, this could be safe since we have control over the invariants and where refcounting occurs. But allowing users to get a `const Tensor&` out of it means that the "internal" Tensor object could be copied outside of your control and you have to be much more careful about correctness, especially since Tensor doesn't seem to know if it's borrowed or not. I saw one of your earlier proposals had Tensor know about this, maybe we should do that?

Or, if the `const Tensor&` conversion is only there to avoid having to rewrite all ops at once, maybe there's a regex/codemod that would allow us to rewrite them more easily and then not allow TensorRef -> const Tensor& conversions? Or, if we manage to eliminate most but not all call sites, we might be able to make it a TensorRef -> Tensor (by value) conversion and be ok with the refcount if somebody does that. Seems like a less dangerous design.

RFC-0010-tensor-ref.md
```diff
@@ -0,0 +1,119 @@
...
+  cycles involving IValue in other ways; in particular, by avoiding
+  use of IValue in Tensor APIs, or by avoiding use of IValue in
+  arguments that are defaulted (defaulting can always be reimplemented
+  with N overloads, one per default argument.  This is not a huge burden
+  in the presence of code generation).
+
+* Instead of trying to force the above implementation of `TensorRef` to
+  be used everywhere, it could instead be a utility class used in
+  limited situations to improve interoperability with code that expects
+  a `const Tensor&` when you don't have a `Tensor` available.  The true,
```
smessmer, line 112: I don't follow why there are two objects - a trivial and a nontrivial one. Can you give some more details?

RFC-0010-tensor-ref.md
ezyang: > But allowing users to get a const Tensor& out of it means that the "internal" Tensor object could be copied outside of your control and you have to be much more careful about correctness

A copy of `const Tensor&` to `Tensor` is OK, because this always induces a refcount bump! Most of the bad situations @swolchok was able to come with required you to have a mutable reference/pointer to the internal Tensor, which we just forbid here.

> Or, if the const Tensor& conversion is only there to avoid having to rewrite all ops at once, maybe there's a regex/codemod that would allow us to rewrite them more easily and then not allow TensorRef -> const Tensor& conversions?

Yeah. I allude to this in the second alternative proposal, where for Intel ABI reasons, it is much better if the long term API doesn't permit conversion to `const Tensor&` (so we can make the class trivial). So ideally we'd get rid of this BC crutch eventually.

RFC-0010-tensor-ref.md
ezyang: The nontrivial TensorRef class is given in this proposal. The trivial class is:

```
class TrivialTensorRef {
  TensorImpl* impl_;
};
```

smessmer: oh I see. I like the trivial one much better from a design point of view ;) But I guess you agree on that point and the non-trivial one only exists for the bc reasons discussed above.

ezyang:
A variation of this has been implemented in https://github.com/pytorch/pytorch/pull/55685

swolchok:
> A variation of this has been implemented in [pytorch/pytorch#55685](https://github.com/pytorch/pytorch/pull/55685)

As with the `TensorRef` in the proposal, `MaybeOwned` is hobbled by the Itanium ABI requirement to pass it by reference always, so it doesn't solve the argument passing problem either.

PR #17, opened 2021-03-17
RFC-0011-InferenceMode

ailzhang:
[Rendered](https://github.com/ailzhang/rfcs/blob/rfc0011/RFC-0011-InferenceMode.md)

### DIFF
--- a/RFC-0011-InferenceMode.md
+++ b/RFC-0011-InferenceMode.md
@@ -35,7 +35,13 @@ Some goals and non-goals:
 * Goal: `InferenceMode` is semantically equivalent to `NoGradMode`,
   except some operations may not be supported.  (In other words, this is
   a partial equivalence: *if* inference mode does not throw an error,
-  then it behaves the same way as no grad mode).
+  then it behaves the same way as no grad mode).  Caveat: this
+  equivalence does not extend to methods that expose private
+  implementation details; esp., `Tensor._is_view` and `Tensor._base`.
+
+* Goal: It should be possible to run code that allocates parameters
+  (tensors with `requires_grad=True`) unchanged inside of an inference
+  mode block.
 
 * Goal: Don't be a global or compile time flag.  This makes
   `InferenceMode` widely applicable as it can still be used in processes
@@ -49,14 +55,15 @@ Some goals and non-goals:
   that `InferenceMode` is safe.
 
 * Non-goal: Make operations on inference tensors fast outside of
-  `InferenceMode`
+  `InferenceMode`; nor, be maximally expressive with inference
+  tensor outside of `InferenceMode`.
 
 * Non-goal: Avoid performance slowdown for view/inplace operations
   outside of `InferenceMode`.  Benchmarking on popular models reveal
   that a slight slowdown on these operations is acceptable; in our
   case, this slowdown will be due to an extra redispatch in these cases.
 
-# User description
+## User description
 
 `InferenceMode` is an RAII guard which can be enabled for a given block
 of code.  Inside inference mode, all newly allocated (non-view) tensors
@@ -64,9 +71,14 @@ are marked as **inference tensors**; these tensors are guaranteed not to
 alias with tensors that may have been saved for backwards (or are
 otherwise making use of version counters--perhaps more accurately,
 you could call these "non version counter tracked tensors").  Inference
-tensors do not have a version counter, and raise an error if you try
-to read their version (e.g., because you saved this tensor for
-backwards.)
+tensors:
+
+* Do not have a version counter.
+* Raise an error if you try to read their version (e.g., because you
+  saved this tensor for backwards.)
+* Raise an error if you try to mutate them into requiring gradients
+  (e.g., directly set `requires_grad=True` or mutate them with a tensor
+  that `requires_grad=True`.)
 
 A non-view tensor is an inference tensor if and only if it was
 allocated during inference mode.  A view tensor is an inference
@@ -78,14 +90,13 @@ guarantees:
 * All operations do not record `grad_fn`, even if their `requires_grad=True`
   (like `NoGradMode`).  This applies for both inference tensors and
   normal tensors (also like `NoGradMode`).
-* View operations do not do view tracking (like `NoGradMode` after
-  PyTorch 1.9).  This applies for both inference tensors and normal
-  tensors.
+* View operations on inference tensors do not do view tracking; views
+  and base inference tensors are indistinguishable.
 * Inplace operations on inference tensors are guaranteed not to do
   a version counter bump (which is equivalent to an atomic increment).
   Inplace operations on normal tensors still do version counter bumps.
 
-# Implementation description
+## Implementation description
 
 **Dispatcher.**  The dispatcher decides what implementation of a kernel
 to call when an operator is invoked.  The set of possible options is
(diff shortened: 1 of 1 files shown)


RFC-0011-InferenceMode.md
```diff
@@ -0,0 +1,138 @@
...
+    //     }
+    //     torch::Tensor c = a.view({4, 1}); // (*)
+    //     If we don't add InplaceOrView to included set, (*) will skip its as_view
+    //     setup entirely, `c` will be a Tensor that is not from Inference mode
+    //     but has potentially wrong view metadata which should be forbidden..
+    //     By going through InplaceOrView kernel, we can throw an error since it
+    //     broke our invariant: "Autograd keys must be in excluded set before
+    //     reaching InplaceOrView kernel".
+```
+3.
```
ezyang, line 133: something here? :)

RFC-0011-InferenceMode.md
```diff
@@ -0,0 +1,138 @@
+Note: a large part of this RFC will become "InferenceMode" documentation once it's finalized.
+
+## Goals:
+- Provide a RAII in C++ and a context manager in Python frontend to switch between inference mode and normal mode, with the following constraints:
```
ezyang, line 4: The Python frontend context manager isn't that important, right? Because this is oriented to performance use cases where you ought to be in C++ only anyway (it's good that it is possible and maybe some time we should add it, but I wouldn't say it's a primary goal)

RFC-0011-InferenceMode.md
```diff
@@ -0,0 +1,138 @@
...
+  - correctness is always guaranteed. (compared to `AutoNonVariableType` which has risks producing silent wrong result.)
+  - performance of infenrence mode should match current existing `AutoNonVariableTypeMode` which is widely used in prod.
```
ezyang, line 6: nit: inference

RFC-0011-InferenceMode.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Goals:
+- Provide a RAII in C++ and a context manager in Python frontend to switch between inference mode and normal mode, with the following constraints:
+  - correctness is always guaranteed. (compared to `AutoNonVariableType` which has risks producing silent wrong result.)
+  - performance of inference mode should match current existing `AutoNonVariableTypeMode` which is widely used in prod.
+  - switching between normal mode and inference mode should be really easy with minimal code change.
+- Make `AutoNonVariableTypeMode` an internal only API, replace all callsites of `AutoNonVariableTypeMode` outside pytorch codebase with the new `InferenceMode`.
+
+## Non-goals:
+- Match the theoretical best inference performance which can be achieved by stripping all autograd related stuff at build time (not flexible).
```
ezyang, line 11: Actually, are we sure about this? If you write code solely in InferenceMode, with no interaction with non-InferenceMode tensors, it seems to me that theoretical best performance should be attainable (since we never have to hit the safety code for the intermixing situation).

RFC-0011-InferenceMode.md
```diff
@@ -0,0 +1,138 @@
...
+- Allowing the most flexible interaction between normal mode and inference mode. Current main use case for inference mode is "either inference or normal" without mixing, so we ban a lot of interactions between two modes to keep the implementation simple.
+
+# Different levels of control over autograd (copied from @Alban)
+The following modes are ranked from slowest to fastest in speed, and from the most flexible to the most restrictive in what users can do.
+
+* Normal Mode: we create the graph for all Tensors that require gradients, always track view and inplace even they don't require gradients.
+* GradMode disabled: we never create the graph, still track all views and inplace. User code always succeeds to properly track gradients.
```
ezyang, line 18: @albanD I kind of want a little more meat here, along the lines of the design principle @apaszke explicated at https://github.com/pytorch/pytorch/pull/12502#pullrequestreview-164417671 

One possible angle here is that `no_grad` is strictly local: the effects of this context manager affect what happens inside the block, but everything outside of the block is still fair game for full autograd support. (This is not as "hard" a design philosophy, but it's my best understanding right now.) Inference mode, on the other hand, is OK with letting the implications of this context escape; its goal is performance, and so if there is no way to implement something fast without affecting stuff outside of the inference mode block, we will just let it leak out.

ezyang:
cc @SsnL, three years later we might actually merge a version of your https://github.com/pytorch/pytorch/pull/12502 haha!

RFC-0011-InferenceMode.md
```diff
@@ -0,0 +1,138 @@
...
+       return at::redispatch::expand(ks & c10::after_InplaceOrView_keyset, self, size, implicit);
+     })();
+     std::function<at::Tensor(const at::Tensor&)> func=nullptr;
+     if (false || !self.unsafeGetTensorImpl()->support_as_strided()) {
+       auto size_vec = size.vec();
+       func = [=](const at::Tensor& input_base) {
+         return input_base.expand(size_vec, implicit);
+       };
+     }
+     auto result = as_view(/* base */ self, /* output */ _tmp, /* is_bw_differentiable */ true, /* is_fw_differentiable */ true, /* view_func */ func, /* creatio
```
ezyang, line 59: copy paste problem

RFC-0011-InferenceMode.md
```diff
@@ -0,0 +1,138 @@
...
+     return result;
+   }
+ ```
+ - **Inference mode** can be turned on when you are sure you don't need any autograd computation. This saves the cost of creating autograd graph and `as_view` / `version_counter` setup compared to the normal mode.
+ - **Inference tensor** is defined as a tensor without Autograd **and** InplaceOrView keys on it.
+ - **Normal tensor** has both Autograd & InplaceOrView keys. This includes both `requires_grad=true` and `requires_grad=false` tensors. (see [Ideal end state] section for more details).
+ - Additional notes:
+   - All Inference tensors are created in inference mode, but not all of the tensors created in inference mode are inference tensors. For example, a view of normal tensor created in inference mode is still a normal tensor (but with special `creation_meta`!).
+   - (Autograd & !InplaceOrView) and (!Autogad & InplaceOrView) are invalid states, we don't have such tensors.
```
ezyang, line 68: Maybe we should have called this Autograd and NotAutograd LOL

RFC-0011-InferenceMode.md
```diff
@@ -0,0 +1,138 @@
...
+
+# Expected Behavior
+## Implementation:
+1. Inference Mode: InplaceOrView not in included, Autograd in excluded
+2. Normal Mode: InplaceOrView in included, Autograd not in excluded
+3. In VariableType kernel, throw an error if input is inference tensor.
+4. In InplaceOrView kernel, throw an error if Autograd keyset is not in excluded set already.
+5. In VariableType kernel, throw an error if input is a view with `NO_VARIABLE_TYPE_VIEW` creation_meta.
+## Behavior
+| Mode          | Input                                    | Op         | Go through Kernels                        | Produced Output                                            |   |   |
```
ezyang, line 78: Is the intention that I can generate all of the entries of the table from the Implementation above?

RFC-0011-InferenceMode.md
ailzhang: Yep agreed, didn't plan to add that until we become stable on the C++ end. Mentioning it here just to make sure it's possible. :D

RFC-0011-InferenceMode.md
ailzhang: oh yes, without the intermixing situation (and maybe some startup cost initializing dispatch table) the performance should be attainable. :D

RFC-0011-InferenceMode.md
ailzhang: you mean autograd tensors and NoAutograd tensors? I like those names but they sounds too related to the GradMode which will be confusing to users :(

RFC-0011-InferenceMode.md
albanD: That good old discussion haha
As mentioned there, this is not really true for gradmode as views created there will lead to some side effects even outside the block.
But I do agree with you that when using this mode, all ops performed outside of it will always work. Which is not true for inference mode.

I am not sure what kind of details you're looking for here?

albanD (approved):
This is now in master!

PR #18, opened 2021-04-05
RFC-0012: Functional lazy traces from XLA to PyTorch

ezyang:
[Rendered](https://github.com/pytorch/rfcs/blob/functional-lazy-traces/RFC-0012-functional-lazy-traces.md)

This is not a complete RFC in the traditional sense, but reflects our current thinking related to the work @asuhan has been doing in torch_xla and how this should affect PyTorch core proper. This doc predates https://github.com/pytorch/xla/issues/2854 but seeing mlir-npcomp pursuing a similar avenue I wanted to get this RFC out there so people outside of FB can have more context about what is going on in XLA internally.

### DIFF
--- a/RFC-0012-functional-lazy-traces.md
+++ b/RFC-0012-functional-lazy-traces.md
@@ -1,0 +1,177 @@
+# Functional lazy traces (from XLA) to PyTorch
+
+**tl;dr** XLA contains a non-XLA specific TorchScript-like IR
+specifically designed for lazy evaluation. Nodes in this IR correspond
+to ATen operations but with mutation complete eliminated at tracing time
+(compared to the incomplete remove mutation pass) and with a distinct
+class per operator node (compared to TorchScript IR which is dynamically
+typed). Alex Suhan makes the case that splitting out this IR from XLA
+and publishing it separately is the fastest way to bring accelerators
+which need lazy tensors online; Edward Yang asks PyTorch composability
+and JIT teams to think about what the best path for this IR to core
+PyTorch is.
+
+## Background
+
+PyTorch XLA allows PyTorch programs to be run by XLA (and most
+importantly, on TPUs) by way of a lazy tensor abstraction. PyTorch
+models executing on the XLA device lazily trace execution until XLA
+requires materialization, at which point the set of operations that were
+run are lowered into XLA HLO IR and then run by XLA itself. Although the
+integration with XLA is imperfect (based on internal work running some
+models on TPU), it remains the best and most comprehensive
+implementation of lazy tensors to date on PyTorch.
+
+Alex Suhan has identified that much of the infrastructure in torch_XLA
+is actually not specific to XLA. In particular, `torch_xla` defines an IR
+*in terms of ATen operators* (not XLA operators; see
+https://github.com/pytorch/xla/tree/master/torch_xla/csrc/ops ), and
+first constructs this IR before lowering into HLO IR. This puts the IR
+in the same conceptual space as TorchScript IR, except:
+
+* It is purely functional (no support for mutation or control flow).
+  XLA’s tracing process keeps tracks of aliases and views so that it can
+  extend the IR at all points where a mutation is visible. This is
+  contrast to the JIT remove mutation pass, which may refuse to remove
+  mutation when it cannot prove it is sound
+  (https://github.com/pytorch/pytorch/issues/39362), and the ONNX
+  functionalization option for tracing, which simply hopes that
+  mutations don’t being visible at other relevant aliases (not always
+  true, see https://github.com/pytorch/pytorch/issues/34538)
+
+* It is specifically designed to be traced every iteration your model is
+  run, so the IR is designed to be more efficient for this use case:
+  e.g., it has separate classes for every ATen operator, rather than a
+  dynamically typed dictionary of attributes; it maintains a hash of an
+  IR node [as it is being constructed](https://github.com/pytorch/xla/blob/08ae1044c2a7e314895f9946104cbe399e096515/torch_xla/csrc/ir.cpp#L149).
+
+* The IR is modestly normalized; for example, optionality is removed in
+  some cases (an optional min value is replaced with an explicit number
+  computed via dtype); some operators are decomposed into a more
+  elementary form (addcmul). The IR also serves as a layer of insulation
+  between PyTorch’s operator bindings (which change frequently with no
+  BC guarantees) and XLA’s lowering (e.g., in
+  https://github.com/pytorch/xla/pull/2692 a change in operator
+  signature resulted in a change in the PyTorch-XLA bridge, but not the
+  IR itself).
+
+Alex has been working on a [refactor of torch_xla](https://github.com/fairinternal/nnc_eager/tree/asuhan/lazy_core_only)
+which splits out this core functionality into its own library that has
+no dependency on TensorFlow/XLA, with the intention of publishing it as
+a library that accelerator vendors can use to bootstrap a lazy tensor
+that records a functional IR which they can then feed to their personal
+IR lowerings. He is also interested in eventually merging this code into
+PyTorch proper.
+
+## How to put this in core?
+
+Although this lazy core functionality is smaller than `torch_xla`, it is
+still not small in an absolute sense (30k lines of C++ code; though
+possibly reducible further), much of it handwritten code for each
+operator XLA supports, but some of it also utility functions that
+arguably should already be available in PyTorch. Furthermore, it has
+some overlap with TorchScript JIT IR (though, as described above, there
+are important differences in goals between XLA’s IR and TorchScript’s
+IR.) This leads to a number of challenges for putting the generic parts
+of XLA in core.
+
+* **How quickly is this needed?** Alex has stated that he would like to
+  ship this API to accelerator partners *within H1 2021*; it is
+  difficult to imagine being able to deliver anything besides a direct
+  drop of this code in that period of time. (Fortunately, it is also OK
+  for this code to live out of repository, for the time being.) How to
+  *move fast* in getting this code into core? (Is a rewrite necessary?
+  Can most code be reused as is?)
+
+* **How to ensure maintainability as new operators are added?** Taken as
+  is, `torch_xla` posits that when a new operator is to be added to the
+  lazy IR, we must define an entirely new C++ class to represent it and
+  write bindings to construct this IR. This is a burden on operator
+  writers; so in practice, operators only get added to the IR when XLA
+  also adds support for the lowering. Should the IR classes be code
+  generated instead? One benefit of the current scheme is that the
+  classes can evolve with a higher degree of backwards compatibility
+  than PyTorch operator definitions usually abide by.
+
+* **How to resolve tech debt from utility classes/functions in XLA that
+  are duplicated in PyTorch?** Did you know torch_xla defines its own
+  C++ type to represent devices? (This is because the “TPU” device is
+  not representable in PyTorch.) There are probably many examples of
+  places in torch_xla where code in PyTorch core could have been
+  adjusted to better accommodate torch_xla’s use cases, but instead the
+  classes/functions were simply reimplemented in XLA to avoid having to
+  get some code into PyTorch core. Should all of this debt be resolved
+  before importing XLA into core? Does it matter?
+
+    * ~~absl dependency~~
+    * lazy_tensors/
+        * Shape (compare with TensorOptions)
+        * PrimitiveType (compare with ScalarType)
+        * Literal (compare with Scalar)
+        * permutation_util.h: utilities for working with permutations
+    * lazy_tensor_core/
+        * data_ops.h: squeeze/unsqueeze shape computation
+        * device.h: reimplementation of device with TPU
+        * helpers.h: type promotion logic, dimension manipulation
+        * shape_builder.h: shape building TensorIterator style
+        * tensor_util.h: stride calculation
+        * torch_util.h: deepcopy, numeric tensor utilities
+
+* **How to ensure maintainability of lazy IR (nb: in memory only, no
+  serialization, no BC) in parallel with TorchScript IR?** One of the
+  biggest maintainability tensions going forward in PyTorch today is
+  juggling the intersection of TorchScript, FX and NFC (all of which
+  have large similarities, but with different constraints.) Lazy IR
+  represents yet another point in the design space where there are a lot
+  of similarities to other existing IRs, but where no other existing
+  solution is easily usable. Can we support adding a fourth blessed IR?
+  How to avoid requiring cross cutting changes making people have to
+  modify four distinct subsystems?
+
+Edward’s opinion:
+
+* Directly shipping XLA’s code as is, is the only plausible route to
+  shipping in the next three months. (It might be possible to prototype
+  within the half a reimplementation of the system from core PyTorch,
+  but it seems exceedingly unlikely that you’d be able to achieve
+  feature parity in this timeframe).
+
+* We need to take a serious look at the backend backwards compatibility
+  problem; it comes up with all of our backend implementors and we need
+  some solution that doesn’t require backend implementors to make
+  changes whenever we do a client side BC change. Theoretically this
+  should be possible (newly added capabilities should simply gracefully
+  fail when sent to a backend that doesn’t support them) but this is
+  difficult to do directly in C++ (code generation would work).
+
+* Personally, I think that people should try to write their lowerings in
+  “finally tagless” style (i.e., bypassing the construction of an
+  intermediate PyTorch IR); however, seeing that preexisting XLA still
+  found it useful to construct an IR corresponding to PyTorch directly,
+  I am not too sure how likely this will actually happen.
+
+* There should be room for more than one IR representation in PyTorch as
+  a whole (think micropass compiler, or bytecode versus graph IR); if we
+  can avoid having to maintain a large apparatus of code for each IR it
+  will make having multiple more tolerable (XLA’s IR is relatively
+  simple, as far as things go, in part because it doesn’t really support
+  mutations on it). So IMO the main problem is socializing this idea
+  with the JIT team (and making sure that the operator story is
+  reasonable).
+
+* I don’t know if a rewrite is right or not, but it should be on the
+  table, particularly for code generating the IR classes. Utilities can
+  probably be refactored incrementally (aka never).
+
+## Appendix: Things that are BC-compat frontend but not backend
+
+One benefit of the hand written class IR is it is more obvious when you
+are breaking BC. In native_functions.yaml most authors think about what
+it means to break client BC, but not backend BC. Note that many of these
+changes are not *morally* BC breaking, but break backends because our
+implementation was bad.
+
+* Renaming overloads
+* Adding a new optional argument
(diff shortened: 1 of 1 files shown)


ezyang:
also cc @wconstab

byronyi:
It seems that torch-xla relies on XLA for shape inference, and PyTorch access the sizes/numel/dim after each op.  If we strip out XLA, this probably need to come from somewhere else.

I also noticed that you added `Meta` device for shape inference. Does lazy tensor depend on shape inference, or in particular, meta tensors implemented in-tree?

asuhan:
@byronyi If we strip out XLA, shape inference can come from PyTorch. Some of the helpers might need to be exposed, but they all exist in core PyTorch, so that's not a fundamental problem - we might need to expose some of them, but they're already written. The XLA inference was just more convenient since we already had to lower to it and could avoid making changes to the core, but the shape inference results are exactly the same and we'll instead use the ones in core PyTorch.

In terms of access to sizes/numel/dim, that works exactly the same with or without XLA, not sure what you mean. What's your concern there?

Finally, lazy tensors don't depend deeply on shape inference or even meta tensors. There's a moderate amount of work to be done for a few operators which currently use static sizes, but nothing fundamental in the design assumes or requires static shapes. We'll still require static ranks, but that should be flexible enough for all practical applications of this.

byronyi:
> Some of the helpers might need to be exposed, but they're already written.

Would appreciate if you could give me a pointer here to core :)

> In terms of access to sizes/numel/dim, that works exactly the same with or without XLA, not sure what you mean. What's your concern there?

I am just wondering if it is possible to lazily evaluate shapes of lazy tensors as well.

> Finally, lazy tensors don't depend deeply on shape inference or even meta tensors. There's a moderate amount of work to be done for a few operators which currently use static sizes, but nothing fundamental in the design assumes or requires static shapes. We'll still require static ranks, but that should be flexible enough for all practical applications of this.

If shapes must be materialized before actually computing the value of lazy tensors, then shape inference seems to be a must to me.

asuhan:
> > Some of the helpers might need to be exposed, but they're already written.
> 
> Would appreciate if you could give me a pointer here to core :)

One such example: https://github.com/pytorch/pytorch/blob/c371542efc31b1abfe6f388042aa3ab0cef935f2/aten/src/ATen/native/ConvUtils.h#L25. They're not centralized, but they're quite easy to find and most interesting / difficult ones tend to be exposed already.

> > In terms of access to sizes/numel/dim, that works exactly the same with or without XLA, not sure what you mean. What's your concern there?
> 
> I am just wondering if it is possible to lazily evaluate shapes of lazy tensors as well.

I think you're conflating two separate aspects here:

1. Shape inference used to validate operations in core PyTorch. That doesn't change in any way because lazy tensors are still ATen tensors and that part still runs before we even hit the vendor back-end. At that stage, sizes are queried directly, just like it happens for CPU or CUDA devices.
2. How to generate code when shapes aren't known when generating code for a graph. That's on the vendor to implement in the compiler back-end, in this tracing infrastructure we'll just have a mode which erases concrete shapes before passing them to the vendor-specific back-end. Once we know that the concrete shapes were valid, we erase them so that we don't trigger re-compilations for the same graph with different concrete sizes, as long as it's structurally the same otherwise. It's then up to the vendor to build a runtime which includes a way to query shapes, hoist / elide bounds checking when redundant etc. For example, for the computation `x + y`, once we know from the higher layer that the shapes were compatible, we just tell the vendor we have `aten::add(x, y)` and the rank of the two tensors. The vendor will pick a representation which includes the shape information in their compiler backend and add runtime calls to query it, to know the upper bound of the loops etc.

> 
> > Finally, lazy tensors don't depend deeply on shape inference or even meta tensors. There's a moderate amount of work to be done for a few operators which currently use static sizes, but nothing fundamental in the design assumes or requires static shapes. We'll still require static ranks, but that should be flexible enough for all practical applications of this.
> 
> If shapes must be materialized before actually computing the value of lazy tensors, then shape inference seems to be a must to me.

I think I've addressed this above. Shapes don't need to be materialized, but they need to be queried with a runtime mechanism which relies on the representation the vendor picks to represent an operator. How the vendor does that is entirely outside of our control since we don't even know how the vendor will represent runtime values.

ezyang:
To add some of my own perspective on top of @asuhan's:

* For size specialized traces (the easy case), asuhan advocates for directly using the size helper functions. This is a good super short term fix. A more mid term fix is to indeed use meta tensors, which offer a *uniform* API for doing shape inference (just convert both input tensors to meta, run the operation on them, and you get a meta tensor that tells you what the output shape should be. Or when https://github.com/pytorch/xla/issues/2871 is a thing we can support codegen facilities to hook into meta computation in a more efficient way). This doesn't save you too much since you have to know a bit about the operators anyway to lower them in a backend, but it is something.
* Based on what @jamesr66a tells me about lazy shape evaluation in FX, you are going to hit some hard walls when a model takes a size somewhere and reads it out and passes it as an integer to some other operation. No matter what you do, XLA-style functional traces in the dispatcher will give you int, and not a symbolic int (because we have C++ code that literally takes int64_t). We're talking about some long term strategies for solving this, but it won't be for a while.

asuhan:
> * Based on what @jamesr66a tells me about lazy shape evaluation in FX, you are going to hit some hard walls when a model takes a size somewhere and reads it out and passes it as an integer to some other operation. No matter what you do, XLA-style functional traces in the dispatcher will give you int, and not a symbolic int (because we have C++ code that literally takes int64_t). We're talking about some long term strategies for solving this, but it won't be for a while.

In the model I'm describing, this would translate to an unavoidable size query runtime call in the vendor backend, by construction it cannot propagate to any dimension size since we erase those. In other words, it'd only be present during shape checks in higher layers, but once we know those pass we erase the sizes and the vendor backend is responsible to emit the proper runtime queries. It's possible the graph would also change, but I'm not sure how frequent that would be in practice.

Tracing is part of the solution and not a full solution. It'll work very well for some models by itself, but we'll definitely need to complement it with TorchScript annotations in parts of the code which are vulnerable to worst aspects of tracing. However, in terms of correctness, I'm pretty sure what I'm describing is sound.

byronyi:
> For size specialized traces (the easy case), asuhan advocates for directly using the size helper functions.

So these size helper functions (for each and every operator in PT core) will be extracted into a central place in a new repo (for lazy tensors)? I am asking this because they seem pretty sparse to me, but maybe that is only because I am not that familiar to the codebase.

> you are going to hit some hard walls when a model takes a size somewhere and reads it out and passes it as an integer to some other operation.

I am actually okay to occasional dynamic shaped/ranked ops as they are probably going to be handled by some fallback logic (i.e. host cpu in XLA case) anyway. But I do hope we could get static/flat trace for a spanning list of uninteresting ops, element-wise, mm, etc., so vendor backends need not to be queried for output shape for each op in the functional lazy trace.

Just as a illustration, for the following wrapper `TensorImpl`:

```
struct XLATensorImpl : public c10::TensorImpl {
  explicit XLATensorImpl(at::Tensor rep)
      : TensorImpl(c10::DispatchKeySet(c10::DispatchKey::XLA), rep.dtype(),
                   c10::Device("xla:0")),
        rep_(std::move(rep)) {
    if (rep_.defined()) {
      set_sizes_and_strides(rep_.sizes(), rep_.strides());
    }
    set_storage_access_should_throw();
  }

  at::IntArrayRef sizes() const override {
    std::cout << "Accessing sizes " << rep_.sizes() << std::endl;
    return rep_.sizes();
  }

  int64_t dim() const override {
    std::cout << "Accessing dim " << rep_.dim() << std::endl;
    return rep_.dim();
  }

  int64_t numel() const override {
    std::cout << "Accessing numel " << rep_.numel() << std::endl;
    return rep_.numel();
  }
};
```

A wrapper-based backend fallback to CPU will gives us the following traces:

```
aten::convolution(Tensor input, Tensor weight, Tensor? bias, int[] stride, int[] padding, int[] dilation, bool transposed, int[] output_padding, int groups) -> (Tensor)
Accessing sizes [4, 10, 24, 24]
Accessing numel 23040
Accessing dim 4
Accessing dim 4
Accessing sizes [4, 10, 24, 24]
Accessing sizes [4, 10, 24, 24]
Accessing numel 23040
Accessing dim 4
Accessing dim 4
aten::max_pool2d_with_indices(Tensor self, int[2] kernel_size, int[2] stride=[], int[2] padding=[0, 0], int[2] dilation=[1, 1], bool ceil_mode=False) -> (Tensor, Tensor)
Accessing sizes [4, 10, 12, 12]
Accessing numel 5760
Accessing dim 4
Accessing dim 4
Accessing sizes [4, 10, 12, 12]
Accessing numel 5760
Accessing dim 4
Accessing dim 4
Accessing sizes [4, 10, 12, 12]
Accessing sizes [4, 10, 12, 12]
Accessing numel 5760
Accessing dim 4
Accessing dim 4
Accessing sizes [4, 10, 12, 12]
Accessing numel 5760
Accessing dim 4
Accessing dim 4
aten::relu(Tensor self) -> (Tensor)
Accessing sizes [4, 10, 12, 12]
Accessing numel 5760
Accessing dim 4
Accessing dim 4
Accessing sizes [4, 10, 12, 12]
Accessing dim 4
```

Ideally the functional lazy traces need not to run the operator on CPU to get sizes and strides. Right now in torch-xla it is handled by XLA shape inference.

asuhan:
> > For size specialized traces (the easy case), asuhan advocates for directly using the size helper functions.
> 
> So these size helper functions (for each and every operator in PT core) will be extracted into a central place in a new repo (for lazy tensors)? I am asking this because they seem pretty sparse to me, but maybe that is only because I am not that familiar to the codebase.

PT/XLA already handles a fraction of operators, not the full set - not all operators occur in real models, especially if we're talking performance-sensitive paths. Further, only some of those have non-trivial shape equations. I've checked most in the past and I remember finding usable helpers for most. I agree with @ezyang that meta-tensors are probably the way to go about this given a longer time horizon, but the situation isn't dire on the short term either.

> I am actually okay to occasional dynamically shaped/ranked ops as they are probably going to be handled by some fallback logic (i.e. host cpu in XLA case) anyway. But I do hope we could get static/flat trace for a spanning list of uninteresting ops, element-wise, mm, etc., so vendor backends need not to be queried for output shape for each op in the functional lazy trace.
> 

The vendor can hoist / elide unnecessary checks for sequences of element-wise operations, for example. It's just a compiler optimization pass in their back-end. We could probably offer some basic, generic tools to assist some of that, but this competes against everything else we want to get done and it's not particularly hard to implement on the vendor end either. But you're right it'd be nice to have it and I'm open-minded about it.

> Ideally the functional lazy traces need not to run the operator on CPU to get sizes and strides. Right now in torch-xla it is sup plied by XLA shape inference.

That would be the idea, you'd never run an operator on CPU just to get sizes and strides. A vendor could choose to punt to CPU whenever a runtime size query would be needed, but that's an implementation quality issue - a good implementation would come with a runtime (native to accelerator) to support such queries on the value representation chosen by a vendor. Think metadata similar to CPU or GPU tensors, but stored in the accelerator memory, nothing prevents that. At the extreme, a pure interpreter for the tensor computation graph can run on the accelerator itself (not saying that'd be a truly good option, just making an informal argument that what I'm describing is feasible).

ezyang:
> So these size helper functions (for each and every operator in PT core) will be extracted into a central place in a new repo (for lazy tensors)? I am asking this because they seem pretty sparse to me, but maybe that is only because I am not that familiar to the codebase.

So, there are two main APIs we are envisioning how to access sizes as we add more support for structured kernels.

First is the direct (but somewhat inefficient) API. Suppose that you are implementing "add" in XLA and you to compute what the output shape should be, you can write this:

```
Tensor add(const Tensor& self, const Tensor& other) {
    auto out_meta = at::add(self.to(kMeta), other.to(kMeta));
    ... out_meta.sizes() ...
}
```

The add call will do all the error checking you need on the inputs, and give you an output type that says what the output sizes, dtype, etc. should be. You can then make use of this information as necessary for your error checking. (NB: the output size isn't *that* useful, for the reason that if you actually are implementing a lowering you will have a far more detailed understanding of what the operator does, but it might still be helpful.

Second is a more pimped up version of @bdhirsh's proposal at https://github.com/pytorch/xla/issues/2871 where we generate the scaffolding that calls into meta, so by the time you are writing lowerings you can assume all error checking has already happened. (We haven't really gotten that far in the design process here).

> Ideally the functional lazy traces need not to run the operator on CPU to get sizes and strides. Right now in torch-xla it is sup plied by XLA shape inference.

Yeah, so structured kernels are trying to make it so that you can easily get the sizes and strides from PyTorch directly, rather than assuming you already have an accurate lowering. Coverage is not very high right now but getting better!

byronyi:
> Yeah, so structured kernels are trying to make it so that you can easily get the sizes and strides from PyTorch directly, rather than assuming you already have an accurate lowering. Coverage is not very high right now but getting better!

Thanks, I was just reading through https://github.com/pytorch/rfcs/blob/rfc-0005/RFC-0005-structured-kernel-definitions.md#goals and that is exactly what I am asking for. We'd love to help on this effort!

ezyang:
There's a nice tracker issue on the subject :) https://github.com/pytorch/pytorch/issues/55070

byronyi:
I see @asuhan pushed to https://github.com/pytorch/pytorch/commit/fafb8ab1f5476418ee121f3dedb6025adc1c9e07 and corresponding changes to torch-xla side: https://github.com/pytorch/xla/tree/asuhan/xla_ltc_plugin. Nice one!

I do see a TODO on shape inference: https://github.com/pytorch/pytorch/blob/lazy_tensor_staging/lazy_tensor_core/lazy_tensor_core/csrc/compiler/node_lowering.h#L19-L20. Do you plan to provide such facilities from core using meta tensors?

asuhan:
@byronyi On the short term I'll just expose / extract the required helpers from core and use them to provide shape inference. Longer term, yes, I really hope we'll be doing that with meta tensors instead - it'd be much nicer.

yzhliu:
@ezyang @asuhan We are working on similar approach to integrate our compiler backend with pytorch. It's great to see the work you are doing. Do you have timeline of when and which part will be moved into pytorch core? Also feel free to let us know if there's anything we can help.

Moreover do you have plan to integrate LazyTensor together with Torchscript so we can add extra layer to scripted modules, as @hzfan mentioned in https://github.com/pytorch/xla/issues/2957#issuecomment-846937949 ?

asuhan:
@yzhliu We intend to put this part into the core, the lazy_tensor_core sub-folder: https://github.com/pytorch/pytorch/tree/lazy_tensor_staging. This was derived from the "upper half" of pytorch/xla - it offers lazy tensor infrastructure which is independent on the actual backend.

We have a working XLA backend: https://github.com/pytorch/xla/tree/asuhan/xla_ltc_plugin, derived from the "lower half" of pytorch/xla and I'm working on a TorchScript backend too - that'll allow us to reuse nvFuser and other TorchScript backends more easily, while providing the graph capture mechanisms and infrastructure.

We have ideas to use TorchScript together with lazy tensors (NB: this is different from TorchScript as a backend mentioned above) to address the caveats - undesired loop unrolling and other similar issues. The idea would be to use it sparingly in problem areas (which have control flow) and still lean on lazy tensors for most of the model, which would maintain usability.

The plan is to become ready to merge lazy_tensor_core to master by July-August. There are several things which need to happen: code base reduction (we can autogenerate most nodes in csrc/ops), removing the last dependencies on absl (we have a handful of uses of StrCat and Span, which we can re-implement) etc. That being said, I'm trying to keep that branch reasonably fresh, 1-2 weeks away from master, until we can finally merge it.

yzhliu:
@asuhan Thanks for sharing the details and the estimated timeline. We'll keep eyes on the project and might bother you folks later :) as we start to make progress.

nunoplopes:
Only saw this RFC now, sorry.
I've been working on a lazy tensor implementation as well. It's fairly small (around 1k LoC + autogen code). I've managed to run a few models from TorchVision & HuggingFace already.
The IR ATM is 1-to-1 with ATen ops. The backend just redispatches op-by-op back to PyTorch. Will hook up with TorchScript next.

The complications I've hit so far:
 - The dtype must be inferred, as the dtype getter is not virtual.
 - it's nice to infer the shape, as otherwise traces are too small in some models.
 - Not all accesses in PyTorch of shape information go through the virtual methods (bugs)
 
For the dtype & shape inference, I wrote a program that executes all ATen ops with different combinations of shapes/dtypes and checks which of my hand-written rules applies. The dispatch wrappers are generated automatically from this information.

My code is available here: https://github.com/nunoplopes/torchy

Would you guys (FB, others?) be interested in chatting about integration, directions, share learned lessons, share data from models running already, etc?

wconstab:
@nunoplopes Thanks for reaching out! We are also continuing on the branch `lazy_tensor_staging` in the subfolder `lazy_tensor_core`, where you can see README and API_GUIDE for more info about the current work.  It would be good to compare notes on the two systems and see if we can combine the best of both.  Also, just curious, what is your use case for lazy tensors?

nunoplopes:
> @nunoplopes Thanks for reaching out! We are also continuing on the branch `lazy_tensor_staging` in the subfolder `lazy_tensor_core`, where you can see README and API_GUIDE for more info about the current work. It would be good to compare notes on the two systems and see if we can combine the best of both. Also, just curious, what is your use case for lazy tensors?

I'm building a JIT compiler for PyTorch. I use lazy tensors to assemble traces and then ship those traces to an off-the-shelf compiler, like TorchScript, for optimization. Traces are cached and reused when seen again.
The goal is to offer eager-mode execution semantics, but with near old-school data-flow performance (XLA & friends).

hzfan:
Do we have a timeline to merge `lazy_tensor_staging` into `main`?

yzhliu:
Hi @ezyang , as mentioned before, we are building features on top of the lazy tensor support, and we noticed the branch is being actively updated. Would you mind share the plan of making the branch officially available to pytorch users?

wconstab:
We are in the process of merging lazy tensor support into pytorch master right now. (You can already see some files landing in `torch/csrc/lazy/...`.  It will probably take us the rest of the year, possibly a little longer, to land the whole thing including our TorchScript backend.  The core interfaces should land sooner, although, we won't consider them 'stable' right away as we are probably going to have to make changes to add support for additional features like distributed training, improve performance, etc.

The actual functionality of lazy tensors won't be enabled 'by default' for users (of cpu/cuda device) this year either.  Right now, lazy is a 'virtual device' meaning you have to move your tensors to 'lazy' (or 'xla') device explicitly, and use other flags to configure the hardware used by the backend.  In the future we'd like to explore making a 'lazy mode' available to existing devices (e.g. cpu, cuda).

hzfan:
@wconstab Could you elaborate a bit more about the TorchScript backend? How does users use TorchScript together with lazy tensor core? Previously I extended ltc to support TorchScript for control flow, and I am super excited to hear that we have official support for it now.

wconstab:
>  elaborate a bit more about the TorchScript backend
It's still WIP, but the idea is to construct a lazy trace of 'functional ATen ops', which are then trivially convertible to torchscript IR.  The lazy-traced torchscript IR is a subset of overall TS IR capabilities, since it is functional, has no controlflow, and has no python classes.  The IR is then fed to a TS GraphExecutor, and used with existing TS passes/backends.

>I extended ltc to support TorchScript for control flow,
Not sure what this means.  We don't support control flow in lazy tracing currently.   (No specific plans to support it either, although, there is some ongoing discussion on the topic.)

PR #21, opened 2021-05-18
RFC-13: Input Descriptor

penguinwu:
This RFC describes the proposal for input descriptor for TorchScript JIT

[Rendered](https://github.com/pytorch/rfcs/blob/input-descriptors/RFC-0013-input-descriptor.md)

### DIFF
--- a/RFC-0013-input-descriptor.md
+++ b/RFC-0013-input-descriptor.md
@@ -1,12 +1,10 @@
 # Motivation
 
-Currently TorchScript supports a generic Tensor type that can be of arbitrary shape, dtype, and layout (i.e., generic tensors). Graphs operating on generic tensors are more dynamic but lack important shape, dtype, layout properties for further optimizations. On the other hand, we observe that many properties of generic tensors can be deduced from those of input tensors.
+Currently TorchScript supports a generic Tensor type that can be of arbitrary shape, dtype, and layout. Graphs operating on generic tensors lack important shape, dtype, layout properties for further optimizations. On the other hand, we also observe that many properties of generic tensors can be inferred from those of input tensors.
 
-Since properties of inputs to a scripted model cannot be obtained by analyzing the model, we propose to augment torch.jit.script API with an optional Input Descriptor parameter that describe tensor input properties (e.g., dtype, device, shape, rank, layout) to the model being scripted. Input descriptor (this RFC) combined with a tensor property propagation engine (not part of this RFC) gives the JIT the ability to propagate these input properties throughout the graphs. In essense it gives the JIT the ability to specialize a graph of generic tensors to a graph of tensors with more known properties that would aid many other downstream optimizations such as AMP, control-flow removal, layout optimization, kernel fusion and codegen, tensor type/shape validation, more effective integration of external optimizers that require static graphs.
+We propose to augment torch.jit.script API with an optional Input Descriptor parameter that describe tensor input properties (e.g., dtype, device, shape, rank, layout) to the model being scripted. Input descriptor (this RFC) combined with a property propagation JIT pass (not included in this RFC) allow the JIT to specialize a graph of generic tensors to a graph of tensors with more known properties without adding much burden to user-level type annotation. This ability would aid many other downstream optimizations (e.g., AMP, control-flow removal, layout optimization, kernel fusion and codegen, tensor type/shape validation) as well enable more effective integrations of external optimizers that require static graphs.
 
 # Design: `torch.jit.script` Input Descriptor
-
-We view `ScriptedModule` as a closure over the internal states of a module object, where states of the module object are captured at the time when `torch.jit.script(model)` is invoked. Let’s make a simplifying assumption of a single entry-point to ScriptedModule (e.g., `forward()`). If we can describe properties of input tensors to scripted functions, then these properties can be propagated to the rest of the IR graph by the JIT.
 
 ## User interface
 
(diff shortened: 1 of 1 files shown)


neginraoof:
Thanks @penguinwu for the RFC. This design would be very useful for ONNX export support of TorchScript modules, and enables export-level verification and inference of input properties. 
Would the descriptor set the data type and shape of an input value?
If so, I think ONNX passes would not need the property propagation pass in order to access input properties. Is this correct?

penguinwu:
@neginraoof The input descriptor can set the dtype and shape (or any other tensor properties) for input if the information is available. However, the input descriptor only describes the properties of input tensors, so if ONNX wants to know the properties of intermediate tensors created inside the model, then it will rely on the tensor property propagation pass to "fill in" tensor properties for those tensors. But if ONNX only needs tensor properties at the entry point to a forward method, then input descriptor would be sufficient.

neginraoof:
@penguinwu Thanks. For now, ONNX exporter uses an ONNX specific logic for shape/type propagation. So input dtype and shape should be sufficient. Do you have any updates on implementation of this feature?

dreiss:
This looks very similar to something we're looking at for the NNAPI converter.  There are two things we'd need in order to adopt this.  Would it be possible to add them?:

1/ For shapes, we need a distinction (on each dimension) between (1) this size is fixed, (2) this size is unknown, but won't vary from batch to batch, so it can be fixed when running the first inference, and (3) this shape can vary from batch to batch.  Would it be possible to extend shape from `List[int]` to `List[SomethingThatCanRepresentThis]`?  We could use negative sizes as sentinels, but that seems hacky.

2/ The NNAPI converter has the ability to semi-transparently convert tensors to NHWC, which gives better performance on many hardware backends.  We need some way for users to indicate that this conversion is required (on a per-tensor basis).  Is that something that "layout" could encompass?  The doc doesn't give examples of how it is meant to be used.

raziel:
Thanks Peng,

From David's comment:

1. So, it seems we need that new symbol: since strings are valid, maybe use char "?" for (2) size is unknown, but will be resolved and fixed.

2. What about having a generic way to express dimension reordering? Would a list of indexes the size of rank with the reordering of the dimensions be enough, worth having, and easy enough for users to specify?

Also:

3. If this descriptor will be used also a documentation, then should it include information like quantization in it?

4. Have you thought about how to integrate other metadata like sparse representation? As long it doesn't prevent adding it in the future that's good, but I'm still curious on your thoughts about it.

penguinwu:
> @penguinwu Thanks. For now, ONNX exporter uses an ONNX specific logic for shape/type propagation. So input dtype and shape should be sufficient. Do you have any updates on implementation of this feature?

Since input descriptor only makes sense when we propagate dtype from input to the rest of the graph and adding the API is relatively straightforward, we decided to get dtype propagation ready first. We were experimenting w/ different designs earlier. @Krovatkin is working on dtype propagation implementation now.

penguinwu:
@raziel @dreiss Answers to your questions/suggestions.

1. >For shapes, we need a distinction (on each dimension) between (1) this size is fixed, (2) this size is unknown, but won't vary from batch to batch, so it can be fixed when running the first inference, and (3) this shape can vary from batch to batch. Would it be possible to extend shape from List[int] to List[SomethingThatCanRepresentThis]? We could use negative sizes as sentinels, but that seems hacky.

I interpret this requirement as how to differentiate between constant length, invariant symbolic length, and dynamic symbolic length. I tweaked a little bit of Raziel's proposal into the following. 

A dimension length can be:
- An non-negative integer number (i.e., representing constant length)
- A string name starting with "?" that represents a symbolic length whose values may vary across runs, e.g., "?i"
- A string name not starting with "?" that represents a symbolic length whose values are only known at runtime but remain invariant across different runs of the same model instance object, e.g., "i"

This proposal use "?i" instead of "?" to represent a dynamic symbolic length because giving dynamic symbolic length a name can express richer semantics. For instance, it can express shapes across two input tensors of a model, `a` of size `["?i", 100, 200]`, `b` of size `["?i", 10]` where the outermost dimensions of both `a` and `b` are dynamic but of the same length (e.g., using the same dynamic batch size).

2.  >The NNAPI converter has the ability to semi-transparently convert tensors to NHWC, which gives better performance on many hardware backends. We need some way for users to indicate that this conversion is required (on a per-tensor basis). Is that something that "layout" could encompass? The doc doesn't give examples of how it is meant to be used.

> We need some way for users to indicate that this conversion is required (on a per-tensor basis).

Let's make sure we are on the same page using an example, consider a model entry point `foo(a: Tensor)` with a single input tensor `a`. Suppose that the captured graph for the model expects `a` to be using a `NHWC` layout. But when we invoke the scripted model, the actual tensor `x` passed in for `a` iss in a different format, say `HCHW`.

I think in order for NNAPI converter to convert `x` from `HCHW` layout to `NHWC` layout, the converter needs two pieces of information, the layout specified for the formal parameter of the model (i.e., layout of `a`) and the layout of the real tensor passed in to `foo`, i.e., `x`. I think the latter can be obtained at runtime. And the conversion could happen during the runtime type-checking when invoking an input descriptor annotated model. In our original design, we only considered reporting errors when the actual arguments do not satisfy the specified input descriptor properties for the corresponding parameter. But I think we could extend the type checking semantics to also include automatic layout conversion (perhaps controlled by a flag if this is not always the desired behavior).

@raziel Regarding your suggestion of "Basically, rather than having an enum to provide the shuffling order of inputs then having a general way to express the input needs to be shuffled around compared with that expected by the model. This obviously assumes that this type of info is all about shuffling dimensions, and not something else altogether.", I feel that Input Descriptor itself is designed to specify properties of input parameters of a model. Directly specifying how to manipulate input is not consistent w/ the type of semantics Input Descriptors are designed for. If we can build the layout transformation semantics into the type checking semantics (as described above) and still satisfy what the kind of behaviors NNAPI converter want to trigger, I feel that would be a cleaner solution. If not, please give me a specific example and what you want to see to happen, then I'll see what we could come up w/

3. > If this descriptor will be used also a documentation, then should it include information like quantization in it?

Input Descriptor is not designed for documentation, it is a specification (contract) on the properties of input parameters to a scripted model. 

4. > Have you thought about how to integrate other metadata like sparse representation? As long it doesn't prevent adding it in the future that's good, but I'm still curious on your thoughts about it.

Right, Input Descriptors can be extended with new properties. Right now, PyTorch supports 3 sparse tensor formats (https://pytorch.org/docs/stable/sparse.html) and I wont be surprised if new sparse formats are added. I could only think of using enum to represent different sparse tensor formats.

raziel:
> This proposal use "?i" instead of "?" to represent a dynamic symbolic length because giving dynamic symbolic length a name can express richer semantics.

That's a good point!

> I think in order for NNAPI converter to convert x from HCHW layout to NHWC layout, the converter needs two pieces of information, the layout specified for the formal parameter of the model (i.e., layout of a) and the layout of the real tensor passed in to foo, i.e., x. I think the latter can be obtained at runtime.

That's right.

> we could extend the type checking semantics to also include automatic layout conversion

That'd be great.

> Directly specifying how to manipulate input is not consistent w/ the type of semantics Input Descriptors are designed for. If we can build the layout transformation semantics into the type checking semantics (as described above) and still satisfy what the kind of behaviors NNAPI converter want to trigger, I feel that would be a cleaner solution.

I think that makes sense. In general there could be a number of things that could be potentially useful for an input descriptor to contain to allow smarter processing. In this case, I was just wondering if there was a way to be able to specify the layout transformation semantics so we didn't have to have hardcoded concepts like "NHWC", which whereas common are specific to a type of tasks/operators.

Right now our only use case is the one David described: NHWC conversion and the main reason is because the backend, in this case, can do something special about it to optimize things further.

> Input Descriptor is not designed for documentation, it is a specification (contract) on the properties of input parameters to a scripted model.
Sorry I didn't express myself correctly. The use case is that inputs (e.g. to Android NNAPI or other backends) may be pre-quantized and that's why we would require quantization parameters (zero point, scale, and possibly whether per-channel quantization is used) for the input descriptors. Granted, quantization would need to match our PyTorch description of a quantized tensor (affine transform, and specific concepts like per-channel which would then be a flag or perhaps an enum --maybe all this should be a QuantizationParams object in the descriptor).

> Input Descriptors can be extended with new properties.
That's right, I was just curious about your thoughts.

dreiss:
> Let's make sure we are on the same page using an example, consider a model entry point foo(a: Tensor) with a single input tensor a. Suppose that the captured graph for the model expects a to be using a NHWC layout. But when we invoke the scripted model, the actual tensor x passed in for a iss in a different format, say HCHW.

For this specific use case, we don't need this level of generality.  We expect all of the inputs passed to the model at runtime to be NCHW, since that is the universal convention within PyTorch.

dreiss:
> The use case is that inputs (e.g. to Android NNAPI or other backends) may be pre-quantized and that's why we would require quantization parameters (zero point, scale, and possibly whether per-channel quantization is used) for the input descriptors.

Oh, yes.  This is also super important.  I didn't notice that this was missing from the original specification.  Thanks for catching it!

penguinwu:
Addressed Raziel and David's comments on:
- Differentiating between static symbolic length and dynamic symbolic length (see subsection on `shape`)
- Augmented the design of `layout` w/ an automatic layout conversion semantics and  (see subsection on `layout` and `type checking and layout conversion); also changed the default value of `layout` from `None` to `NCHW`
- Added `quantized` property (see subsection on `quantized`) but detailed design not yet added

raziel:
Thanks @penguinwu 

> * Added `quantized` property (see subsection on `quantized`) but detailed design not yet added

We can work on it when you're available. It should be fairly straightforward, and in parallel we're also aligning on the naming of the terminology for our quantization APIs anyway, which should be consistent here.

garymm:
Does the fact that this hasn't been merged mean that the design is not agreed upon and no further work is happening on input descriptor?

penguinwu:
Given that there is currently no adoption of the input descriptor design -- the intended use cases (e.g., selective build and input specialization) have adopted other approaches -- I will close this pull request.

(We may resurrect it if future use cases surface)

jgong5:
> Given that there is currently no adoption of the input descriptor design -- the intended use cases (e.g., selective build and input specialization) have adopted other approaches -- I will close this pull request.

May I know what "other approaches" you are referring to? I know that PyTorch is primarily doing tracing with concrete values in various features, e.g., torch.jit.trace, torch.fx. define-by-run quantization.

PR #23, opened 2021-07-01
RFC for profile directed typing using Monkeytype tool in TorchScript

nikithamalgifb:


RFC-0012-profile-directed-typing.md
```diff
@@ -0,0 +1,295 @@
+## Overview
+
+Missing/incorrect type annotations is a common reason for failed model scripting. For a long time, the process of mitigating missing/incorrect type annotations has been through trial and error, ie. by repeatedly hitting the wall of various type-checking errors generated from torch.jit.script and fixing them one by one. This is both inefficient and frustrating.
```
ansley, line 3: I would prefer "A common reason for failed model scripting is missing or incorrect type annotations", because then the first part of the sentence is sort of a "build up" to the solution we're proposing

RFC-0012-profile-directed-typing.md
```diff
@@ -0,0 +1,295 @@
...
+
+Stand-alone Tool with Human Intervention to Permanently Annotate Source Code
+
+Another option we thought about is to create a stand-alone tool that is based on MonkeyType to help users permanently annotate their source code, instead of only generating one-time type annotations on the fly as we proposed in this approach.
+
+The additional benefit is that stand-alone tool could optionally ask for user’s opinion on whether annotated types match their expectation and make revision based on user input.
+
+However, using a stand-alone tool has a few down-sides:
+
+* MonkeyType is unaware of TorchScript limitation of expressing nn.Module types and would blindly apply nn.Module and its subclasses as type annotation. Though legal in Python, they are not accepted in TorchScript. We need to configure MonkeyType to ensure it never annotates module types.
```
ansley, line 291:
```suggestion
* MonkeyType is unaware of the TorchScript limitation of expressing nn.Module types and would blindly apply nn.Module and its subclasses as type annotations. Though legal in Python, they are not accepted in TorchScript. We need to configure MonkeyType to ensure it never annotates module types.
```

RFC-0012-profile-directed-typing.md
```diff
@@ -0,0 +1,295 @@
...
+* TorchScript has sophisticated type inference for return values of functions/methods, empirically it is more than enough to infer all return types of functions. Having MonkeyType annotate return types is more likely to cause type conflicts rather than helping with compilation.
```
ansley, line 292:
```suggestion
* TorchScript has a sophisticated type inference for return values that is more than enough to infer all return types of functions. Having MonkeyType annotate return types is more likely to cause type conflicts than helping with compilation.
```

RFC-0012-profile-directed-typing.md
```diff
@@ -0,0 +1,295 @@
...
+* User code may reference third-party libraries that user installed in system-wide locations. If we have a stand-alone tool that profiles types and applies derived type annotations to them, then user environment is polluted. If type annotations are not applied to them, then TorchScript would fail.
```
ansley, line 293:
```suggestion
* User code may reference third-party libraries that user has installed in system-wide locations. If we have a stand-alone tool that profiles types and applies derived type annotations, then the user environment would be polluted. On the other hand, if the type annotations were not applied, then TorchScript would fail.
```

RFC-0012-profile-directed-typing.md
```diff
@@ -0,0 +1,295 @@
...
+* Using a stand-alone tool is a less stream-lined experience compared to type inference embedded in torch.jit.script calls because users need to create another program that drives compilation target function.
```
ansley, line 294: *streamlined

RFC-0012-profile-directed-typing.md
```diff
@@ -0,0 +1,295 @@
...
+* Intuitively I feel creating such stand-alone tool belongs and quip it with human intervention possibility is more like an enhancement to MonkeyType rather than being a part of PyTorch.
```
ansley, line 295: opinion, please delete

ansley (changes requested):

ansley:
The content of this is good, but I saw some repeated mistakes that I would like to address before publishing this:
1. Mismatched subjects. For example, either the third-person singular or the imperative is fine when describing what code does/should do, but please choose one and stick with it.
2. Too many filler words, e.g. "very", "a lot", etc. I think it's good style to limit filler words in professional writing
3. Incorrect use of commas
4. Run-on sentences (related to the above point)
5. Links to FB-internal sources
6. Lack of Markdown formatting. In particular, please put backticks around all code and format links
7. Sentences vs. sentence fragments. Either one is fine for bullet points, but choose one and stick with it

Note that, per my comments, I did NOT flag all instances of the above mistakes. It would have been too messy if I did that, and I didn't want to overload you by requesting the same changes repeatedly. Please keep your eye out for the above mistakes when you're fixing the other things I requested.

Overall good work! Thanks again for writing this up

PR #24, opened 2021-07-31
RFC-0015-Tracer-module-scope-extension

BowenBao:
[rendered](https://github.com/BowenBao/rfcs/blob/master/RFC-0015-Tracer-module-scope-extension.md)

### DIFF
--- a/RFC-0015-Tracer-module-scope-extension.md
+++ b/RFC-0015-Tracer-module-scope-extension.md
@@ -1,0 +1,74 @@
+## Summary
+Extending `TracingState.push_scope` and `TracingState.pop_scope` with function overload to take module inputs, attributes, and outputs as additional arguments. Enalbing more information of the module call to be recorded when `recording_scopes` is turned on.
+
+## Motivation
+A particular `Scope` object is created for each `nn.Module` `forward` call when `recording_scopes` is turned on. Each node within the `forward` call is annotated with that `Scope` object. This enables ONNX export to group nodes based on scope of original `nn.Module`.
+
+However, with current approach, the information of inputs, outputs, and attributes of `nn.Module` call is missing. Graph analysis can be done to figure out which `Value` should be input/output, based on the scope location of their creation and use. Yet it is no longer possible to retrieve the input/output order or input/output name. For module attributes, it is even worse as some attributes are only used in module initialization, and not visible/retrievable in traced graphs.
+
+The goals are
+
+* Goal: enable jit (ONNX) passes to retrieve inputs/outputs/attributes information for each `Scope`.
+
+* Goal: only affect code path of ONNX export when this feature is specifically turned on.
+
+## User description
+
+Users are not aware of this implementation detail.
+
+## Implementation description
+
+**TracingState.**
+```c++
+  py::class_<TracingState, std::shared_ptr<TracingState>>(
+      m, "TracingState", py::dynamic_attr())
+      .def(
+          "push_scope",
+          [](TracingState& s,
+             const std::string& scope_name,
+             const py::tuple& input_tuple,
+             const py::dict& input_dict,
+             const py::dict& attributes) {
+            // ...
+          })
+      .def(
+          "pop_scope",
+          [](TracingState& s, const IValue& output) {
+            // Retrieve Value* in IR Graph, given IValue.
+            auto v = s.getValue(output);
+            // ...
+          })
+```
+
+**model.py**
+```python
+    # push scope
+    from typing import get_type_hints
+    annotations = get_type_hints(type(self))
+    base_m_annotations = get_type_hints(torch.nn.Module)
+    if annotations is not None and base_m_annotations is not None:
+        [annotations.pop(k, None) for k in base_m_annotations.keys()]
+    else:
+        annotations = {}
+    attrs = {k : getattr(self, k) for k in annotations.keys()}
+    tracing_state.push_scope(name, input, kwargs, attrs)
+
+    # ...
+
+    # pop scope
+    tracing_state.pop_scope(result)
+```
+
+**recording_scopes**. Currently decided by `recording_scopes = torch.jit._trace._trace_module_map is not None`. Perhaps need to add new flag to enable the new feature.
+
+**Information propagation**. With new API, a mapping between `Scope` of module, and `Value*` of module inputs/outputs/attributes can be established. Now what's left is to preserve and propagate this info throughout jit (onnx) passes to be used.
+
+One option could be extending `Graph` class with below info for each `Scope`.
+```c++
+value_list inputs;
+std::unordered_map<std::string, Value*> kwargs;
+std::unordered_map<std::string, IValue> attributes;
+value_list outputs;
+```
+`Node::destroy()`, `Graph::freeNode()`, `Value::replaceAllUsesWith` etc must be modified accordingly to ensure the values recorded above are also correctly maintained. This approach needs some work for inputs/outputs, however it should be straightforward for attributes.
+


RFC-0015-Tracer-module-scope-extension.md
```diff
@@ -0,0 +1,74 @@
+## Summary
+Extending `TracingState.push_scope` and `TracingState.pop_scope` with function overload to take module inputs, attributes, and outputs as additional arguments. Enalbing more information of the module call to be recorded when `recording_scopes` is turned on.
```
jiafatom, line 2: nit: Enabling

RFC-0015-Tracer-module-scope-extension.md
```diff
@@ -0,0 +1,74 @@
...
+
+* Goal: only affect code path of ONNX export when this feature is specifically turned on.
+
+## User description
+
+Users are not aware of this implementation detail.
+
+## Implementation description
+
+**TracingState.**
```
garymm, line 21: Please link to the file that would be modified.

RFC-0015-Tracer-module-scope-extension.md
```diff
@@ -0,0 +1,74 @@
...
+      .def(
+          "pop_scope",
+          [](TracingState& s, const IValue& output) {
+            // Retrieve Value* in IR Graph, given IValue.
+            auto v = s.getValue(output);
+            // ...
+          })
+```
+
+**model.py**
```
garymm, line 43: ditto

RFC-0015-Tracer-module-scope-extension.md
```diff
@@ -0,0 +1,74 @@
...
+    attrs = {k : getattr(self, k) for k in annotations.keys()}
+    tracing_state.push_scope(name, input, kwargs, attrs)
+
+    # ...
+
+    # pop scope
+    tracing_state.pop_scope(result)
+```
+
+**recording_scopes**. Currently decided by `recording_scopes = torch.jit._trace._trace_module_map is not None`. Perhaps need to add new flag to enable the new feature.
```
garymm, line 62: Please link to where this code is currently.

ezyang:
Let's chat about this in VC some time; module scopes seem very similar to capabilities that FX provides, and I'm curious if we want to try hitting this with a more radical approach (e.g., rewriting ONNX to be an FX pass rather than a TorchScript trace based pass)

garymm:
@ezyang I've been told you have been invited to our regular meeting on 09-02 and we can discuss then.
However this scope feature is needed in the short term and we don't want to block it on rewriting the entire ONNX export.
Could we discuss the FX topic independently of moving forward with this for now?

ezyang:
I finally had a chance to think about this proposal. I think I am primarily skeptical about the proposal is to add new attributes to Graph to record the scope information.  If you're going to maintain a mapping from string scope name to extra args/kwargs, there's no reason this couldn't just have been done from Python, where when tracing you maintain this mapping. You push a unique scope string and then stash in the dictionary the extra args/kwargs you want to record. An added benefit about this change is that you don't have to deal with serializing Python arguments to IValues; you can record them verbatim (and store arbitrarily complicated stuff, like the nn.Module itself). WDYT? We can talk about this more on Tuesday too.

RFC-0015-Tracer-module-scope-extension.md
```diff
@@ -0,0 +1,74 @@
...
+**TracingState.**
+```c++
+  py::class_<TracingState, std::shared_ptr<TracingState>>(
+      m, "TracingState", py::dynamic_attr())
+      .def(
+          "push_scope",
+          [](TracingState& s,
+             const std::string& scope_name,
+             const py::tuple& input_tuple,
+             const py::dict& input_dict,
```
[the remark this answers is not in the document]
BowenBao: (1) correct. *args and **kwargs
(2) likely not, since args is tuple.
(3) tracing cannot handle optional input. We observe what's provided in sample input for trace call.

RFC-0015-Tracer-module-scope-extension.md
```diff
@@ -0,0 +1,74 @@
...
+**model.py**
+```python
+    # push scope
+    from typing import get_type_hints
+    annotations = get_type_hints(type(self))
+    base_m_annotations = get_type_hints(torch.nn.Module)
+    if annotations is not None and base_m_annotations is not None:
+        [annotations.pop(k, None) for k in base_m_annotations.keys()]
+    else:
+        annotations = {}
```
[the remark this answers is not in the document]
BowenBao: Edge case handling if `annotations` is override by users for the module class. Consider no annotation and abort in this case.

BowenBao:
@ezyang the Python side idea is appealing for its simplicity. We are considering something like this for the attribute side especially it they are scalars or python numbers. It is not immediately clear to me how to achieve the goal of identifying and preserving mapping between python args and IR Values with this approach. 

Consider a simplified example for this purpose.
```python
class Layer(torch.nn.Module):
    def forward(self, x, y):
        return ...

class Model(torch.nn.Module):
    def __init__(self):
        ...
        self.layer = Layer()

    def forward(self, input):
        x = ...  # produced from some module
        y = ...  # produced from some module
        return self.layer(x, y)
```

```
# IR snippet:
%x = ...
%y = ...
%out.1 = ONNX::layer_op(...)  # part of Layer module
%out.2 = ONNX::layer_op2(...)  # part of Layer module
%out.3 = ONNX::layer_op3(...)  # part of Layer module
return (%out.3)
```

Current IR does not explicitly show `%x`, `%y` are the Layer module inputs, and `%out.3` is the Layer module output. Graph analysis can tell if Value is produced from or used by outer scope, but the original inputs/outputs ordering cannot be preserved. For our proposed approach we do not intend to preserve the IValue itself. The IValue is passed in through `push_scope`/`pop_scope` only to look up its IR Value from the tracer stash.

ezyang:
Hmm. I'm not sure I entirely understand, but would it be helpful if you could retrieve the IR values *while* you were tracing (or maybe, some sort of token that you could use later to correspond to the graph)? Then you'd be able to trace Python arguments.

ezyang:
We discussed this in our meeting and the key constraint from ONNX's side is that they want to be able to access the module scope information from C++. I'm not too sure what to do about this, TBH. It will be helpful to know what kinds of optimizations would actually make use of this information.

ezyang:
Hmm... what if we add some `__torch_function__` style functionality so that you can interpose on the nn.Module call at the time you are tracing, and immediately do the `custom_domain` replacement, so that is the only thing that ever shows up in the trace?

albanD:
>  what if we add some __torch_function__ style functionality so that you can interpose on the nn.Module call at the time you are tracing

Module hooks?
Feel like global module hook (that fire for every Module that run) is what you're looking for in this case.

BowenBao:
> Hmm... what if we add some `__torch_function__` style functionality so that you can interpose on the nn.Module call at the time you are tracing, and immediately do the `custom_domain` replacement, so that is the only thing that ever shows up in the trace?

We'd also like to still trace the inner subgraph of the modules, since that will appear as the onnx local function subgraph definition.

ezyang:
I guess what is giving me a lot of trouble here (and why the responses take a while), is that I don't really want to add a new fundamental, cross-cutting change to TorchScript IR with only one motivating use case for it. I'd rather take some much more circuitous route; for example, the original proposal with doing things in Python; what if you did the same thing, but transmitted the arguments all the way to C++ so they were accessible from your TorchScript passes? It's still hacky, but it feels a lot better.

BowenBao:
@ezyang That's a valid point, I guess we can explore not changing TorchScript IR, and try to manage things in Python. 
Instead of then `push_scope`/`pop_scope` proposal, we could possibly depend on `_get_value_trace` api from `TracingState` to retrieve the mapping of `IValue <-> Value`.

For the global module hook idea above, do you have some pointers or example snippets that we could give a try?

ezyang:
^- @albanD

BowenBao:
@albanD @ezyang friendly ping for response. Is [this](https://github.com/pytorch/pytorch/blob/5950240bdf4fe0734ea47180dfbd9b69c7091fa9/torch/nn/modules/module.py#L52) the api for 

> global module hook (that fire for every Module that run)


In the link, the comment states

> The input contains only the positional arguments given to the module.
    Keyword arguments won't be passed to the hooks and only to the ``forward``.

Is it possible to relax the constraint to support kwargs as well?

ezyang:
> Is it possible to relax the constraint to support kwargs as well?

We should; iirc the only reason it wasn't done was because it was BC breaking

albanD:
@BowenBao sorry I missed this message.
Yes this is the right API.
Adding kwargs has been discussed and is indeed a bit tricky due to BC but we're ok with adding a boolean flag when registering the hook to specify if the kwargs should be passed or not: https://github.com/pytorch/pytorch/pull/61606 (we could remove the lazy module part of this and add just the hook improvement).

PR #26, opened 2021-08-09
Should Pytorch Provide Performance Hints

msaroufim:
@chauhang @HamidShojanazeri @gdankel @dzhulgakov @jerryzh168

### DIFF
--- a/RFC-0012-PerformanceHints.md
+++ b/RFC-0012-PerformanceHints.md
@@ -1,0 +1,100 @@
+# Should Pytorch Provide Performance Hints?
+
+Pytorch prides itself on being a user centric library where historically most of its users have been researchers. However, a growing number of developers are interested in deploying Pytorch with the best possible out of box throughput while meeting some hard latency constraint in inference.
+
+Unfortunately, getting these speedups while possible requires lots of domain knowledge and without lots of Googling and talking to people it's difficult to figure out how to proceed next. I went through this experience trying to get the best possible performance on a BERT model and summarized some of my observations below.
+
+The goal of this RFC is to write down optimizations people have in their heads in code.
+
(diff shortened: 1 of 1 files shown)


albanD:
Hey,
Thanks for writing all of this down!

Most of these suggestions seem to be related to extension libraries or specific models that don't live into core.
Could you give some details of where you would expect this kind of package to live?  Would it be a third party repo that would get imported in the user code and would start monitoring third party library usage and also provide some APIs for the user to monitor their code?

msaroufim:
Hi @albanD it depends on the suggestion but my thinking is 

* Pytorch Distributed handles whether you are using it correctly
* HuggingFace tells you that you're overpadding or you could use a slower model
* Pytorch or Pytorch Profileer tells you your accelerator utilization is low
* Torchserve tells you that you can use quantization but maybe even Pytorch?
* torch.device could hold more optimization related data
* Pytorch tells you to set number of threads to 1 on CPU and tells you that batch size needs to be powers of 2

albanD:
That sounds like a lot of different things the users needs to know about to get that information?
In particular because most of these checks wouldn't be enabled by default (because they would be too spammy or too expensive to run).
Wouldn't a third party repo that would be a one place for all of these be easier to use from a user perspective?

msaroufim:
Yeah that's precisely the problem that motivated me to write this. But now that you mention it, it seems like the natural place for a lot of these ideas is either in Pytorch Profiler where people are opting in to get verbose advice

msaroufim:
Closed to discuss this internally first

soumith:
I think the precise thing to discuss here is how to maintain and manage these heuristics that we throw out as part of the PyTorch Profiler (if we accept this PR).

The social aspects of it, the robustness of the suggestions, how one accepts or rejects a suggestion as part of someone contributing it in a PR, if these suggestions are related to non-core libraries, how one handles mainstream ecosystem libraries vs obscure ecosystem libraries, etc. are all important, challenging and need some proposed thoughts.

PR #27, opened 2021-08-26
RFC-0016: Masked reductions and normalizations

cpuhrsch:
This RFC Discusses semantics and implementation details of masked reduction and normalization operators.

[Rendered](https://github.com/pytorch/rfcs/blob/8cb9ce7fe84724099138dc281080c74ad1bc2cca/RFC-0016-Masked-reductions-and-normalizations.md)

### DIFF
--- a/RFC-0016-Masked-reductions-and-normalizations.md
+++ b/RFC-0016-Masked-reductions-and-normalizations.md
@@ -1,0 +1,259 @@
+### [RFC] Masked reductions and normalizations
+
+
+#### Purpose of this document
+
+Discuss semantics and implementation details of masked reduction and normalization operators. 
+
+
+#### State of the world
+
+We currently have a few operations in PyTorch that take masks as inputs. nn.Transformer and related operators and operators masked_scatter, masked_select and masked_fill.
+
+
+#### Project goals
+
+Extend reduction and normalization semantics to apply only to a subset of values as determined by a separate masked input. This aims to close multiple issues and the related user use cases and feature requests.
+
+**Motivation**
+
+
+
+* There is some explicit user demand for reductions with masked semantics for sparse Tensors
+    * [[feature request] Sparse (hybrid sparse-dense) output option for topk, min, max ](https://github.com/pytorch/pytorch/issues/34951)
+    * [Feature request: sparse matrix max(axis)](https://github.com/pytorch/pytorch/issues/4247) 
+    * [Implemented torch.sparse.max(sparse_tensor, dim)](https://github.com/pytorch/pytorch/pull/59340)
+* [RFC-0004: Adding fill value property to PyTorch sparse tensors](https://github.com/pytorch/rfcs/pull/8) is blocked on discussions around masked semantics.
+* There are also demands for masked normalizations for dense Tensors and some prior art that shows performance improvements.
+    * [Add masked_softmax to speed up masking in multihead attention](https://github.com/pytorch/pytorch/pull/48441)
+    * [Fuse softmax and masking in MultiheadAttention](https://github.com/pytorch/pytorch/issues/44945)
+    * [[FR] Safe softmax](https://github.com/pytorch/pytorch/issues/55056)   
+
+**Existing reduction operators**
+
+Reductions are operators that remove a dimension from a Tensor. They apply a binary operation across a list of Tensor slices and aggregate the result. This is different from normalizations that return the same shape with all dimensions intact.
+
+Dense ([docs](https://pytorch.org/docs/master/torch.html#reduction-ops))
+
+
+
+* argmax
+* argmin
+* amax
+* amin
+* aminmax
+* all
+* any
+* max
+* min
+* dist
+* logsumexp
+* mean
+* median
+* nanmedian
+* mode
+* norm
+* nansum
+* prod
+* quantile
+* nanquantile
+* std
+* std_mean
+* sum
+* unique
+* unique_consecutive
+* var
+* var_mean
+* count_nonzero
+
+Sparse COO ([docs](https://pytorch.org/docs/master/sparse.html#torch-functions-specific-to-sparse-tensors))
+
+
+
+* sum
+
+Sparse CSR: None
+
+**Existing normalization operators**
+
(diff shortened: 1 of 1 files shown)


ezyang:
> does this mean you'd prefer to see this released and packaged out of tree first before considering inclusion in the core?

Not necessarily; I'm referring to this part of the spec:

> Indeed the best way to describe the behavior is to implement it. Please note that this is only meant to describe semantics and is not an actual implementation.

wouldn't be a long step to have an executable specification that people can play around with.

mruberry:
Since the nan* reductions, like [nansum](https://pytorch.org/docs/master/generated/torch.nansum.html?highlight=nan#torch.nansum), are existing masked reductions we should be sure the semantics are equivalent. This proposal just allows the mask to be specified directly rather than by value. Supporting more general value-based masking might be interesting in the future, too.

cc @heitorschueroff

cpuhrsch:
@ezyang - agreed, I'm wondering whether or when we should create an out-of-tree Python-only prototype for a MaskedTensor.

@mruberry - you can always get a value-based mask by e.g. ```masked_sum(input, ~(input != 4))``` or ```masked_sum(input, ~(input != input))``` for nan.

pearu:
> `masked_sum(input, ~(input != input))` for nan.

Nit: `masked_sum(input, input == input)` would work for the nan case as well.

ezyang:
> agreed, I'm wondering whether or when we should create an out-of-tree Python-only prototype for a MaskedTensor.

If it's just one person, probably sticking it in a colab is good enough. Multiple people wanting to work on the semantics ~> put it in GitHub somewhere.

RFC-0016-Masked-reductions-and-normalizations.md
```diff
@@ -0,0 +1,259 @@
...
+
+* softmax
+* log_softmax
+* norm / native_norm
+
+Sparse CSR: None
+
+**Operator constraints and general signature - Reductions**
+
+**Input types:** A mask is a boolean tensor. It accompanies a dense Tensor of the same shape. If an entry is True the corresponding element at the same index in the paired dense Tensor is a "valid" value. If it is False it is not. "valid" here means that this value is meant to be included in the computation and otherwise is meant to be ignored. This matches the semantics of masked_scatter, masked_select and masked_fill.
```
pearu, line 108: This definition of mask mismatches the definition of [numpy.ma mask](https://numpy.org/doc/stable/reference/maskedarray.generic.html#what-is-a-masked-array):
```
When an element of the mask is False, the corresponding element of the associated array is valid
and is said to be unmasked. When an element of the mask is True, the corresponding element of
the associated array is said to be masked (invalid).
```
Possible solutions:
1. Adjust the mask definition to match the one in numpy.ma
2. Rename "mask" to "valid"
3. Document the mismatch with numpy.ma

I would vote for 1 because the mismatch with the three non-arithmetic ops `masked_scatter`, ... would not be as bad as the mismatch with arithmetic ops of more widely used numpy.ma module, IMHO.

pearu (commented):
I have two suggestions
- re mask definition that mismatches with the one from numpy.ma
- re API of masked operations to match the requirements of ReductionOpInfo

RFC-0016-Masked-reductions-and-normalizations.md
pearu: I will redraw my vote for option 1 and suggest defining:
```
A mask is a boolean tensor of selection.
```

cpuhrsch: I think the current definition of mask adheres to ```A mask is a boolean tensor of selection.``` if I understand you correctly?

Interestingly enough our [MHA implementation](https://pytorch.org/docs/master/generated/torch.nn.MultiheadAttention.html#torch.nn.MultiheadAttention) uses True to indicate an element is meant to be ignored.

But like you mention the other ops use the inverse of this definition.

pearu: > I think the current definition of mask adheres to `A mask is a boolean tensor of selection.` if I understand you correctly?

yes

> Interestingly enough our [MHA implementation](https://pytorch.org/docs/master/generated/torch.nn.MultiheadAttention.html#torch.nn.MultiheadAttention) uses True to indicate an element is meant to be ignored.

I guess it boils down to not assuming that "mask" is a uniquely defined concept and the exact meaning of mask must be documented in functions docstrings.

> But like you mention the other ops use the inverse of this definition.

In addition, https://numpy.org/doc/stable/reference/generated/numpy.sum.html uses `where` argument as a "mask of selection".

RFC-0016-Masked-reductions-and-normalizations.md
```diff
@@ -0,0 +1,259 @@
...
+**Operator coverage:** As a first step we should implement the basics: sum, prod, mean, amin, amax.
+
+**Semantics:** Indeed the best way to describe the behavior is to implement it. Please note that this is only meant to describe semantics and is not an actual implementation. It also implies the gradients for these operations. Whether an input or mask is backed by a dense or compressed storage layout has no influence on the semantics of these operations. More concretely, unspecified values for sparse layouts are always zero in this context. If the mask is None the behavior of the operator is equivalent to the unmasked variant.
+
+**Input layouts:** If an input is passed a sparse Tensor, it will return a sparse output. At first the layout of the mask and input must match. If the input is a sparse Tensor, the layout of the mask must be too. This is to ease kernel development and might quickly change, if there is user demand. However, the user can always convert a mask to sparse or dense before invocation.
+
+
+```
+def masked_sum(input, dim, keepdim, dtype, mask):
+    return torch.sum(input * mask, dim, keepdim, dtype=dtype)
```
cpuhrsch, line 121: NOTE: This actually requires masked_fill, because we have no guarantee that any masked values of input are valid.
```
>>> torch.tensor([float('inf')]) * torch.tensor([False])
tensor([nan])
>>> torch.tensor([float('inf')]).masked_fill(~torch.tensor([False]), 0)
tensor([0.])
```

RFC-0016-Masked-reductions-and-normalizations.md
cpuhrsch: Based on our offline discussion with Ralf I think it's safe to go for mask semantics that adhere to those described in this RFC.

PR #28, opened 2021-09-14
RFC-0017-Autograd-subgraphs-PT-ONNX-Export

shubhambhokare1:
[Rendered](https://github.com/pytorch/rfcs/blob/880aa3464d14ef29471e4a3c05f06d0c1cb43b02/RFC-0017-Autograd-subgraphs-PT-ONNX-Export.md)

### DIFF
--- a/RFC-0017-Autograd-subgraphs-PT-ONNX-Export.md
+++ b/RFC-0017-Autograd-subgraphs-PT-ONNX-Export.md
@@ -1,0 +1,118 @@
+## Summary
+Expanding the traced graph resulting from a `torch.autograd.function` in order to enable inline export from PyTorch to ONNX. Expansion of graph of such functions to happen only when the PT-ONNX exporter passes the relevant flag.
+
+## Motivation
+Currently, the PT-ONNX exporter enables the export of a `torch.autograd.function` using two different methods.
+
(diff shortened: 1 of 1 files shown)


albanD:
Are you supposed to be able to do training on the exported model?

RFC-0017-Autograd-subgraphs-PT-ONNX-Export.md
```diff
@@ -0,0 +1,118 @@
...
+    bool is_inplace,
+    bool unpack_output) {
+  if (!jit::tracer::isTracing()) {
+    return;
+  }
+  // ...
+
+  for (const auto i : c10::irange(num_outputs)) {
+    PyObject* obj = PyTuple_GET_ITEM(output_objects, i);
+    if (THPVariable_Check(obj) && !getTracingState()->expand_autograd_op) {
```
BowenBao, line 107: `&& !getTracingState()->expand_autograd_op` This is the only place of change in this code section right?

shubhambhokare1:
@albanD yes, the intention is to be able to training on the exported model. However this change is mainly for terms of inference and exporting models with autograd functions that have complex structures that prevent us from writing symbolics for them

albanD:
That's interesting, because removing the custom Function will change the behavior of the backward pass. So we should forbid training if we remove the custom Function right?

PR #32, opened 2021-11-20
RFC-0020/0021/0022 RFCs for Pipeline Parallelism

jamesr66a:
This PR consists of three RFCs:

* RFC-0020 Pipeline Parallelism Strategic Plan
* RFC-0021 Pipeline Parallelism Technical Approach Proposal
* RFC-0022 Model Partitioning in Pipeline Parallelism Proposal

### DIFF
--- a/RFC-0020-Distributed-Pipeline-Parallelism.md
+++ b/RFC-0020-Distributed-Pipeline-Parallelism.md
@@ -1,0 +1,209 @@
+# [RFC] Ceci n'est pas pipeline parallelism (Pipeline Parallelism 2021Q4/2022 Plan)
+
+This is an RFC for the strategic plan for further developing pipeline parallelism in PyTorch. **We invite our users and partners to comment on this plan and the corresponding technical plan** to help us develop the best APIs for PP in PyTorch.
+
+Goal: *Provide a flexible, composable, and reconfigurable interface for pipeline parallelism in PyTorch that allows scaling a wide variety of PyTorch models on a wide variety of hardware configurations*.
+
+## Motivation
+
+* Pipeline parallelism (PP) is used as a lower-communication-volume technique for model parallelism. It is especially applicable when data must be transmitted across comparatively slower interconnects.
+* Several research-oriented frameworks exist that implement PP (e.g. fairscale, megatron, deepspeed), but we would like to provide a production-quality implementation and support contract for PP.
+* The existing PP implementation in PyTorch (`torch.distributed.pipeline.sync`) only supports intra-host pipeline parallelism across GPUs and does not support techniques like 1F1B scheduling. We can deliver inter-host pipelining and other features.
+* Ultimately, we want to use this body of work as a driving force for research in delivering both performance AND usability of parallelism paradigms. We invite developers and researchers to participate in the design and development of this project.
+
+## Stage 1: Requirements Gathering (2021Q4)
+
+We have spent a good amount of time this calendar quarter researching the user requirements and systems research directions of pipeline parallelism and will continue to do so going forward. **We invite additional comments to fill in details we have not captured here**, if any.
+
+### Prior Work
+
+The research literature[1-12] has a rich body of work. This includes:
+
+* Synchronous vs. Asynchronous pipeline parallelism.
+    * Synchronous pipeline parallelism where a mini-batch is split into micro batches and the pipeline is filled and drained, blocking until the mini-batch is completed. This is the typical use case we are designing for
+    * Asynchronous pipeline parallelism that keeps the pipeline continually occupied. Various clever techniques such as weight stashing and weight prediction have been proposed to address the consistency issues from the "locking" nature of SGD in these cases. These techniques may introduce additional design concerns in a pipeline parallelism API.
+* Pipeline scheduling, where the execution order of `forward` or `backward` micro-batches follows a specified policy. The infrastructure for implementing these schedules can be an important consideration for the design on a PP API.
+    * Fill-drain schedule, where all forward micro-batches are run to completion before all backward micro-batches are run and parameter updates are applied.
+    * 1F1B schedule, where `backward` micro-batches are triggered by the last pipeline stage and stages ensure that they alternate between running `forward` and `backward` micro-batches at steady-state. This helps to reduce the amount of state stored on each pipeline stage.
+    * More, including interleaved 1F1B and new research schedules.
+
+### Key Stakeholders
+
+This section is meant to capture key users/researchers who would benefit from such a pipeline parallelism API. **We invite additional comments to fill in users/researchers who would benefit from this API and would like to see their requirements satisfied**.
+
+#### P0: HF Transformers
+
+HF transformers [wants to](https://github.com/huggingface/transformers/issues/13690) incorporate 3d parallelism including Pipeline Parallelism, however the [current PyTorch implementation](https://github.com/pytorch/pytorch/blob/9f4e004abd8c5d11fc23f4ab705328cb9b4050bb/torch/distributed/pipeline/sync/pipe.py#L220) has limitations that we should address (Px is a priority, with lower x being higher priority. We assigned these priorities based on a) user need and b) implementation time/complexity, but we can adjust them based on user feedback):
+
+* Frontend limitations:
+    * **P0**: Cannot pass arbitrary data types between pipeline stages
+    * **P0**: Unclear composability in 3d parallelism scheme (data, pipeline, model parallel)
+    * **P1**: User needs to rewrite their model as an `nn.Sequential` instance
+* Backend Limitations:
+    * **P(-1)**: No cross-host support for PT pipeline parallelism API
+    * **P0**: No support off-the-shelf schedules (1F1B or interleaving)
+    * **P1**: No support arbitrary programmable schedules
+* Non-requirements:
+    * Composability with ZeRO-2/3 is not required. Theoretically possible, but reportedly will not give any perf gain.
+* Success Criteria:
+    * **to be determined**: Feedback on this would be appreciated
+
+### Prior Implementations and Proposed Approach
+
+An analysis of prior implementations and a proposed technical approach for pipeline parallelism can be seen in [[RFC] Distributed Pipeline Parallel Training Technical Approach](https://github.com/pytorch/rfcs/blob/master/RFC-0021-Distributed-Pipeline-Parallel-Technical.md). In this document, we further split execution into stages and correlate those to the PyTorch external release schedule.
+
+## Stage 2: Ship prototype synchronous multi-node pipeline parallelism (torchgpipe-style) (1.11 Prototype Release)
+
+### P(-1): Implement cross-host support for pipeline parallelism
+
+Existing approaches that support this (in no particular order):
+
+* Fairscale [experimental distributed pipeline parallelism](https://github.com/facebookresearch/fairscale/tree/main/fairscale/experimental/nn/distributed_pipeline)
+* Sagemaker [model parallelism](https://arxiv.org/abs/2111.05972)
+* [DeepSpeed pipeline parallelism](https://www.deepspeed.ai/tutorials/pipeline/)
+* [OneFlow](https://github.com/Oneflow-Inc/oneflow)
+
+Proposed approach short-list: (all approaches can be seen in [[RFC] Distributed Pipeline Parallel Training Technical Approach](https://github.com/pytorch/rfcs/blob/master/RFC-0021-Distributed-Pipeline-Parallel-Technical.md)
+
+1. Selected approach: "Approach 3 with Modifications"
+    * Inherit RemoteModule + torchpipe-based implementation from fairscale [experimental distributed pipeline parallelism](https://github.com/facebookresearch/fairscale/tree/main/fairscale/experimental/nn/distributed_pipeline).
+    * Switch autograd off of distributed autograd and onto manual handling of autograd in the pipeline, to facilitate implementing schedules (e.g. 1F1B)
+    * Abstract the runtime for each RemoteModule to allow for programming in execution schedules
+    * Switch from using DistributedLoss to having a loss callback to facilitate the last pipeline stage calling the loss locally rather than relying on the training loop to calculate the loss via RPC and call distributed autograd. This will be necessary with arbitrary schedules.
+
+### P0: Implement support for passing arbitrary data-types between pipeline stages
+
+Existing approaches that support this (in no particular order):
+
+* Some amount of [support](https://github.com/pytorch/pytorch/issues/53952) in existing PT implementation
+
+Proposed approach short-list:
+
+1. Hopefully should just work out of the box with the RPC API, but need to keep it in mind.
+
+### P0: 1.11 Prototype Release and out-of-tree demo on HF Transformers
+
+* Release API as prototype in the 1.11 release to facilitate gathering feedback
+* Validation: Out-of-tree demo on HF transformers repo - hack it together to get it to work and pull out work items to improve the API to remove places where code edits are needed
+* 1.11 Release Dates
+    * Feature submission: 11/30 EOD
+    * Branch cut 1/31/2022
+
+
+
+### **P0**: Support off-the-shelf schedules (1F1B or interleaving)
+
+Existing approaches that support this (in no particular order):
+
+* Megatron hardcoded schedules: [1f1b](https://github.com/NVIDIA/Megatron-LM/blob/5ac5571ba0265af4c491ee0af1508ca7589450c6/megatron/schedules.py#L517), [interleaved](https://github.com/NVIDIA/Megatron-LM/blob/5ac5571ba0265af4c491ee0af1508ca7589450c6/megatron/schedules.py#L187)
+
+Proposed approach short-list:
+
+1. "Approach 3 with Modifications"
+    * Once manual autograd is handled, and we abstract the workers, we can implement 1F1B or interleaved 1F1B using that infrastructure.
+
+### P0: Composability of PP with TP and DP (3d Parallelism)
+
+Existing approaches that support this (in no particular order):
+
+* `torch.distributed` APIs via module wrapper composition
+* [DeepSpeed](https://www.deepspeed.ai/tutorials/pipeline/)
+
+Proposed approach short-list:
+
+1. DistributedDataParallel wraps PipelineParallel API which operates on upcoming ShardedTensors
+2. Unified programming model (Stage 5)
+
+## Stage 3: Figure out how to reconcile local pipeline vs. distributed pipeline (2022H1)
+
+The existing approaches live in different corners of a 2-dimensional space with axes on **single-driver vs. actors** and **local vs. distributed**.
+
+
+|	|single-driver	|actors	|
+|---	|---	|---	|
+|local	|torchgpipe/fairscale Pipe/distributed.sync.Pipe	|	|
+|distributed	|	|Fairscale distributed_pipeline, DeepSpeed, Megatron-LM	|
+
+### Design Speculation
+
+We can interpolate the missing spaces:
+
+* **single-driver, distributed**: “macro SIMD” style distributed execution. I believe this is actually what was envisioned in @pritamdamania87’s [RFC](https://github.com/pytorch/pytorch/issues/44827) with the `torch.distributed.pipeline_sync` API. The current `distributed.sync.Pipe` API is a fork of the `torchgpipe` implementation (transitively forked in `fairscale`), which is hard-coded for single-node execution issuing commands via CUDA streams (or a fake CPU stream stand-in they implemented)
+* **actors, local**: We can take the event-driven approach taken in fairscale’s `distributed_pipeline` and extend that to having worker processes/threads that both a) feed a corresponding CUDA device and b) feed data through to the successor in the pipeline. This is sort-of already done by the `torchgpipe` lineage of implementations which use [worker threads](https://github.com/pytorch/pytorch/blob/master/torch/distributed/pipeline/sync/worker.py) that run the actual forward computation but still have a central coordinating thread issuing each of those workers commands nonetheless. Potentially if done in a multi-process setting, this could lead to higher performance (need to measure).
+
+I believe the way to go in the future may be to consolidate on actors for both local and distributed. This may represent lower complexity than the torchgpipe-style execution (at least when I think about it) and can avoid issues with a single driver process being a bottleneck (as evidenced by the fact that `torchgpipe` already uses threads for speed).
+
+
+## Stage 4: Generalize pipeline parallelism interface to allow for more coverage of different techniques in the literature (e.g. async, scheduling, auto-partitioning, composition with tensor parallelism) (2022, OSS releases 1.11-1.15)
+
+### P1: Pipeline parallelism without `nn.Sequential` rewrite
+
+Existing approaches/proposals that support this (in no particular order):
+
+* Sagemaker [model parallelism](https://drive.google.com/file/d/1N2eo5Yr_QOw0EtKv-MYBDWKvyRYxKv2o/view)
+* @zdevito's [sequential-free splitting approach](https://colab.research.google.com/drive/1lGg2NqlvDwVmvBqejzni2yTmYE9rxfdr?usp=sharing)
+* [OneFlow](https://github.com/Oneflow-Inc/oneflow)
+* [[RFC] Model Partitioning in Pipeline Parallelism](https://github.com/pytorch/rfcs/blob/master/RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md)
+
+Proposed approach short-list:
+
+1. [[RFC] Model Partitioning in Pipeline Parallelism](https://github.com/pytorch/rfcs/blob/master/RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md)
+2. @zdevito's [sequential-free splitting approach](https://colab.research.google.com/drive/1lGg2NqlvDwVmvBqejzni2yTmYE9rxfdr?usp=sharing)
+3. Construct a pipeline parallelism API that uses a different approach, such as the one used in SageMaker model parallelism. This introduces trade-offs elsewhere, such as in support for schedules/the requirement for an optimization pass to be applied to implement "true" pipeline parallelism.
+
+These approaches can be composed on top of an existing API that takes an `nn.Sequential`. We may consider in the future to develop a "v2" API that is centered more natively around non-`nn.Sequential` models using technologies from Sagemaker, OneFlow, or other research developments.
+
+### P1: Support arbitrary programmable schedules (e.g. fill-drain, 1F1B, interleaved 1F1B) 
+
+Existing approaches that support this (in no particular order):
+
+* DeepSpeed [PipeSchedule](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/pipe/schedule.py) is an instruction format that allows customizing the order in which forward/backward jobs on different stages should be executed.
+
+Proposed approach short-list:
+
+1. Programmable instruction stream + interpreter à la [PipeSchedule](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/pipe/schedule.py). This should be enabled by the infrastructural work done in Stage 2.
+
+### P2: Asynchronous Pipeline Parallelism - Mechanics of Asynchronous Training Loop
+
+**Call for Stakeholders**: Do you have a project that would benefit from Asynchronous Pipeline Parallelism in PyTorch? Please comment on the RFC and we can incorporate your requirements.
+
+* async training is like a self-perpetuating engine v.s. a synchronous procedure call as is typical in Python. How do we bridge these two? What would the Pythonic experience for async look like?
+
+### P2: Asynchronous Pipeline Parallelism - Weight stashing
+
+**Call for Stakeholders**: Do you have a project that would benefit from Asynchronous Pipeline Parallelism in PyTorch? Please comment on the RFC and we can incorporate your requirements.
+
+* [Parametrization](https://pytorch.org/tutorials/intermediate/parametrizations.html) as an approach?
+
+### P2: Asynchronous Pipeline Parallelism - Double-Buffered Weight Stashing
+
+**Call for Stakeholders**: Do you have a project that would benefit from Asynchronous Pipeline Parallelism in PyTorch? Please comment on the RFC and we can incorporate your requirements.
+
+* [Parametrization](https://pytorch.org/tutorials/intermediate/parametrizations.html) as an approach?
+
+### P2: Asynchronous Pipeline Parallelism - Weight Prediction
+
+**Call for Stakeholders**: Do you have a project that would benefit from Asynchronous Pipeline Parallelism in PyTorch? Please comment on the RFC and we can incorporate your requirements.
+
+* [Parametrization](https://pytorch.org/tutorials/intermediate/parametrizations.html) as an approach?
+
+## Stage 5: Integrate into Unified Programming Models Research (2022?)
+
+Going into the future, we would like to develop theory and implementation for a unified distributed, parallel programming model that brings together all of data parallel, model parallel, pipeline parallel, expert parallel, and more. Various ideas are floating around, including building on top of the Actor model (as in Ray, OneFlow, etc) or extending the MPI-style SPMD model to support spatial parallelism like pipeline parallelism and predicated expert parallelism. Hopefully, this pipeline parallelism project will help to inform us on the correct model here and we can publish our findings in the future.
+
+
+## References
+
+1. Efficient and Robust Parallel DNN Training through Model Parallelism on Multi-GPU Platform https://arxiv.org/abs/1809.02839
+2. ElasticPipe: An Efficient and Dynamic Model-Parallel Solution to DNN Training https://dl.acm.org/doi/10.1145/3322795.3331463
+3. XPipe: Efficient Pipeline Model Parallelism for Multi-GPU DNN Training https://arxiv.org/abs/1911.04610
+4. PipeDream: Fast and Efficient Pipeline Parallel DNN Training https://arxiv.org/abs/1806.03377
+5. GPipe: Efficient Training of Giant Neural Networks using Pipeline Parallelism https://arxiv.org/abs/1811.06965
+6. torchgpipe: On-the-fly Pipeline Parallelism for Training Giant Models https://arxiv.org/abs/2004.09910
+7. Pipelined Backpropagation at Scale: Training Large Models without Batches https://arxiv.org/abs/2003.11666
+8. Memory-Efficient Pipeline-Parallel DNN Training https://arxiv.org/abs/2006.09503
+9. Efficient Large-Scale Language Model Training on GPU Clusters https://arxiv.org/abs/2104.04473
+10. Performance analysis of a pipelined backpropagation parallel algorithm https://ieeexplore.ieee.org/document/286892
+11. PipeMare: Asynchronous Pipeline Parallel DNN Training https://arxiv.org/abs/1910.05124
+12. Scaling Language Model Training to a Trillion Parameters Using Megatron
+ https://developer.nvidia.com/blog/scaling-language-model-training-to-a-trillion-parameters-using-megatron/
--- a/RFC-0021-Distributed-Pipeline-Parallel-Technical.md
+++ b/RFC-0021-Distributed-Pipeline-Parallel-Technical.md
@@ -1,0 +1,620 @@
+# [RFC] Distributed Pipeline Parallel Training Technical Approach
+
+## Background - PyTorch Training Loop
+
+PyTorch does not vend a standard training loop abstraction. As a result, the training process for a PyTorch model consists of free-form Python code. An example of a standard training loop in PyTorch might look something like this (borrowed from the PyTorch transfer learning [tutorial](https://pytorch.org/tutorials/beginner/transfer_learning_tutorial.html)):
+
+```
+def train_model(model, criterion, optimizer, scheduler, num_epochs=25):
+    since = time.time()
+
+    best_model_wts = copy.deepcopy(model.state_dict())
+    best_acc = 0.0
+
+    for epoch in range(num_epochs):
+        print('Epoch {}/{}'.format(epoch, num_epochs - 1))
+        print('-' * 10)
+
+        # Each epoch has a training and validation phase
+        for phase in ['train', 'val']:
+            if phase == 'train':
+                model.train()  # Set model to training mode
+            else:
+                model.eval()   # Set model to evaluate mode
+
+            running_loss = 0.0
+            running_corrects = 0
+
+            # Iterate over data.
+            for inputs, labels in dataloaders[phase]:
+                inputs = inputs.to(device)
+                labels = labels.to(device)
+
+                # zero the parameter gradients
+                optimizer.zero_grad()
+
+                # forward
+                # track history if only in train
+                with torch.set_grad_enabled(phase == 'train'):
+                    outputs = model(inputs)
+                    _, preds = torch.max(outputs, 1)
+                    loss = criterion(outputs, labels)
+
+                    # backward + optimize only if in training phase
+                    if phase == 'train':
+                        loss.backward()
+                        optimizer.step()
+
+                # statistics
+                running_loss += loss.item() * inputs.size(0)
+                running_corrects += torch.sum(preds == labels.data)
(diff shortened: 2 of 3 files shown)


RFC-0021-Distributed-Pipeline-Parallel-Technical.md
```diff
@@ -0,0 +1,620 @@
...
+
+* Approach 1: SPMD with Predicated Training Loop and message passing
+* Approach 2: RPC with RemoteModule and torchgpipe-style single coordinator (@pritamdamania87 RFC)
+    * Single-coordinator
+        * CUDA command buffer analogy
+    * Continuation passing
+* Approach 3: MPMD with RemoteModule and message passing (fairscale experimental)
+* Approach 4: MPMD with a custom interpreter/instruction format and message passing (DeepSpeed)
+* Approach 5: RPC with remote modules and generalized Module-server architecture (SageMaker)
+* Approach 6: SPMD with Program capture/JIT compilation and message passing (OneFlow)
```
yf225, line 578: I am curious how quickly we can prototype some of the approaches on typical transformer models such as GPT3-13B or GPT3-175B, and then evaluate them on both UX and performance aspects.

IMO from pure UX point of view we might want to preserve **Python training loop** and **free-form (no sequential) authoring**, otherwise the adoption barrier might be so high that users would just not try it.

I even feel that maximizing UX at a small cost of perf is fine, because user can always find a way (and hire help) to hyper-optimize their training pipeline later if needed.

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
```diff
@@ -0,0 +1,620 @@
...
+
+PP_RANK 0 drives the process by scheduling instances of the training loop function (a UDF annotated by `@smp.step`): two for each micro-batch (one for forward, one for backward). PP_RANK 0 can implement different “schedules” by dispatching these `(micro-batch, phase)` tuples in a given order. The orders that they present are:
+
+
+* Simple pipeline (aka GPipe fill-drain). This is implemented by having PP_RANK 0 dispatch the `phase=forward` tuples for each micro-batch in sequence. Then, dispatching the `phase=backward` tuples for each micro-batch in sequence.
+* “interleaved” pipeline (**NB**: this is not the same as the *interleaved 1F1B* from Narayanan, 2021). PP_RANK 0 will schedule `phase=forward` jobs and opportunistically schedule `phase=backward` jobs *as soon as the forward pass for that micro-batch is done*.
+
+![Module-server request-response execution in SageMaker pipeline parallelism](https://i.imgur.com/y9MZJ3b.png)
+
+**NOTE**: The schedules here do not necessarily run stage in a given order on each stage. Network latency and other affects may change the order of when micro-batches are executed.
```
yf225, line 508: I feel that non-determinism in scheduling is probably okay as long as we can prove it's correct and we can reduce bubble and memory usage as much as possible.

Particularly for memory, one thing that's tricky these days is to reason through the memory usage behavior and to optimize it (e.g. to fit a bigger batch size). Since in general forward accumulates activation tensor memory and backward releases activation tensor memory, iiuc under this scheme, doing forward twice might cause OOM while using a strict 1F1B schedule wouldn't. In that case I think having a deterministic (or at least a formally reasonable) schedule is still better.

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
jamesr66a: Thanks @yf225! Indeed, we are also hearing feedback from other users that the front-end restrictions will be prohibitive. I think this week I will likely keep iterating on both strategy (prioritize non-`Sequential` models) and technical approaches that address the front-end issue and update this RFC.

deepakn94:
Thanks for this, @jamesr66a! It is quite detailed; it will take me a few more days to give it the full attention it deserves.

Some very quick high-level thoughts:
- I prefer figuring out a clean API to support pipeline parallelism for any model, rather than first trying to support `torch.nn.Sequential`. Things like skip connections are pretty common, so figuring out how to support these now rather than later seems like time well spent.
- I also think supporting this with as few code changes as possible to the main training loop is good, similar to `torch.distributed.DistributedDataParallel`. Of course, some things will probably need to change given that every rank is not loading input data, etc.
- I like the idea of hiding the specifics of what happens in a given batch (e.g., how many microbatches are in a batch, how these microbatches are scheduled, etc.) behind some API. IMO this achieves a clean separation of concerns: users can use free-form Python for the per-batch processing (as before), while hiding parallelization-strategy-specific implementation details behind an API. And the API can provide some way for users to override the schedule if they like. I wonder if it's possible to support asynchronous pipelining schemes using this approach.

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+DAPPLE combines DDP with pipeline parallelism to form a bigger search space and use a more efficient schedule if possible. Their partition approach is a DP-based algorithm, it first tries to find the “pivotal” stage, then optimize the overall latency,  the “overall latency” optimization here tries to reduce the bubbles in the pipeline as small as possible.
+
+
+
+## Proposing: Automated Model Partitioning in PyTorch Pipeline Parallelism
+
+Given the existing work and their limitations, we introduce a set of APIs that’s flexible enough for future improvements and intuitive to use. We expose a single API `create_balanced_partition`, to take the model and do the partition under the hood.
+
+```
+def create_balanced_partition(model: nn.Module,
```
deepakn94, line 54: This could perhaps be a bit more expressive if you also considered data parallelism degree. So given a total number of devices, you figure out the data parallelism and pipeline parallelism degrees as well.

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+                              devices: List[RemoteDevice],           
+                              *sample_input) -> List[RemoteModule]:
+ #device = torch.device("cuda")):
+ # if model itself if larger than a single cuda device memory, 
+ # should we allow the user to profile the model on cpu?
+ # Decision: probably no, as the characteristics are different
```
deepakn94, line 60: Yeah, I think you still want to do profiling on the same target GPU type. Perhaps you can stream execution per operator / inner module in order to keep memory footprint low?

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+
+Note that this will be the only API we expose to the user, it returns a partitioned model which already been transferred to the corresponding devices. This allow us to iterate on the underlying implementation and experiment more efficient partition algorithms in the future.
+
+## FX compatibility
+
+Pipeline parallelism auto partition capability needs more advanced knowledge in order to accurately divide a model into a balanced partition set.  In order to achieve auto partitioning with the most balanced approach, we need to get the model execution graph and try to split the model base on the graph. Using torch.fx can give us a helpful graph representation in order for us to do more advanced partition with extensive analysis with tracing and profiling. 
+
+The model that passed in should be fx compatible in order to generate the partitions with the most accurate estimation. What models does not have fx compatibility currently:
```
deepakn94, line 87: Probably want to collect memory profiles as well (especially for large models that don't fit in a single accelerator).

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+cost(m) = time_weight * execution_time(m) 
+            + memory_weight * memory_cost(m)
+
+memory_cost(m) = parameter cost * optimizer_multiplier 
+                    + activation_cost
+```
+
+where `memory_weight + time_weight = 1`. How to weigh between time and memory cost? Undecided, a heuristic number `time_weight=0.6`
+
+* @mrshenli: regarding balancing time between memory, I kind of feel we should prioritize time over memory. Because people usually use Pipeline parallel to accelerate training, and the training speed the ultimate goal. Balanced execution time has a directly impact on total pipeline makespan (suppose one phase is slower than others, then all other devices will need to wait for that phase). If the above assumption is correct, it looks like we should first try to balance time, and only try to balance memory when a shard of the optimally time-balanced model cannot fit in some devices.
```
deepakn94, line 179: People use pipeline parallelism often to train large models like GPT-3, which don't fit on a single accelerator. I think of memory_cost as a constraint rather than an objective (minimize time-per-iteration while ensuring that any worker's memory footprint does not exceed its memory capacity).

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+    SGD        2--3           (momentum_buffer)
+    Adam       4--5           exp_avg, exp_avg_sq, (max_exp_avg_sq)
+    Adadelta   4              square_avg, acc_delta
+    Adagrad    3              sum
+    RMSprop    3--5           square_avg, (momentum_buffer), (grad_avg)
+    =========  =============  =========================================
+```
+
+
+Since we assigned the cost to each top-level module, we can do the partition, basic partition function will be like the below:
```
deepakn94, line 201: Instead of hardcoding these constants, can we not run a few training iterations with the relevant optimizer?

RFC-0020-Distributed-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,209 @@
...
+* [DeepSpeed](https://www.deepspeed.ai/tutorials/pipeline/)
+
+Proposed approach short-list:
+
+1. DistributedDataParallel wraps PipelineParallel API which operates on upcoming ShardedTensors
+2. Unified programming model (Stage 5)
+
+## Stage 3: Figure out how to reconcile local pipeline vs. distributed pipeline (2022H1)
+
+The existing approaches live in different corners of a 2-dimensional space with axes on **single-driver vs. actors** and **local vs. distributed**.
```
deepakn94, line 119: What does `local` and `distributed` mean here?

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
```diff
@@ -0,0 +1,620 @@
...
+|Approach 4	|multi	|interp	|seq	|local?	|manual	|local	|async*	|X	|?	|	|
+|Approach 5	|single*	|py	|FF	|local?	|manual	|local?	|sync?	|X	|X	|Schedules?	|
+|Approach 6	|multi	|interp	|FF	|local?	|manual? (graph?)	|local?	|async?	|X	|X	|	|
+
+### Decision - Approach 3 with Modifications
+
+After deliberation, we want to build the API with the least complexity, at least initially. We will modify/build the API in FairScale experimental with a few modifications:
+
+
+* (**DA5**) Rather than using torchgpipe-style virtual dependencies in the distributed autograd graph, we want each stage to manually handle running `forward` and `backward` stages (the latter by explicitly calling `torch.autograd.backward()`). This will give easier and finer-grained control over the execution schedule of the pipeline
```
deepakn94, line 594: Can an example of what the main training loop looks like with these modifications be added?

And also, how is training launched? Some example commands for each required rank / host would be useful.

RFC-0020-Distributed-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Stage 4: Generalize pipeline parallelism interface to allow for more coverage of different techniques in the literature (e.g. async, scheduling, auto-partitioning, composition with tensor parallelism) (2022, OSS releases 1.11-1.15)
+
+### P1: Pipeline parallelism without `nn.Sequential` rewrite
+
+Existing approaches/proposals that support this (in no particular order):
+
+* Sagemaker [model parallelism](https://drive.google.com/file/d/1N2eo5Yr_QOw0EtKv-MYBDWKvyRYxKv2o/view)
+* @zdevito's [sequential-free splitting approach](https://colab.research.google.com/drive/1lGg2NqlvDwVmvBqejzni2yTmYE9rxfdr?usp=sharing)
+* [OneFlow](https://github.com/Oneflow-Inc/oneflow)
```
stas00, line 145:
```suggestion
* [OneFlow](https://github.com/Oneflow-Inc/oneflow)
* [Varuna](gttps://github.com/microsoft/varuna) / [paper](https://arxiv.org/abs/2111.04007)
```

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+Partition Algorithm: For partition_method = “parameters”, DeepSpeed use the mechanism of counting the layer’s parameters size, and do a binary search (find the smallest weight of the heaviest partition) to find a balanced partition.
+
+***PipeDream***
+More on asynchronous training pipeline planning, like parameter server approaches, the algorithm itself also uses Dynamic Programming, but the states is highly correlated with the communication and synchronization costs, which is not a good choice for us (unless we consider async training in the future). 
+
+***DAPPLE***
+
+DAPPLE combines DDP with pipeline parallelism to form a bigger search space and use a more efficient schedule if possible. Their partition approach is a DP-based algorithm, it first tries to find the “pivotal” stage, then optimize the overall latency,  the “overall latency” optimization here tries to reduce the bubbles in the pipeline as small as possible.
+
+
```
stas00, line 47: Here too Varuna that I linked too earlier provides a partitioning mechanism using cutPoints. see their paper : https://arxiv.org/abs/2111.04007 for details.

stas00 (commented):
Thank you so much for working on this!!! So happy to see this RFC.

I left a few small suggestions if you will find those useful.

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Proposing: Automated Model Partitioning in PyTorch Pipeline Parallelism
+
+Given the existing work and their limitations, we introduce a set of APIs that’s flexible enough for future improvements and intuitive to use. We expose a single API `create_balanced_partition`, to take the model and do the partition under the hood.
```
thomasw21, line 51: Should we be able to "override" the partitioning, ie similarly to feed to a regex like rule?

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+
+Our approach here starts with an approach similar to fairscale, which only do top-level partition, but it could be improved further by exploring more partition approaches, i.e. with operation level partitioning by torch.fx, we could partition the model in a more even manner, which further improves the efficiency. Since we only expose a simple API and return the partitioned model with stages assigned to the devices, the internal implementation can be improved as we explore more ideas. The plan for partition algorithm exploration:
+
+* Use fairscale/gpipe's [Block Partition of Sequences](https://arxiv.org/pdf/1308.2452.pdf) as our first step
+* See if we can apply [Fiduccia-Mattheyses Heuristic](https://en.wikipedia.org/wiki/Fiduccia%E2%80%93Mattheyses_algorithm) to partition the graph and how the performance compare with the default one
+* See if we can do operation-level tracing, and partition the model into several balanced `torch.fx.GraphModule` instead of using the original module architecture
```
thomasw21, line 150: Would that mean that we lose all methods linked to the original class? For example `transformers` fit a certain interface which allows them to use a `generate` method. If one uses `torch.fx.GraphModule` it loses all methods linked to it's original module?

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+ # returned a fx-based model with partition results applied (assigned to devices already)
+ return partitioned_model
+```
+
+
+Note that this will be the only API we expose to the user, it returns a partitioned model which already been transferred to the corresponding devices. This allow us to iterate on the underlying implementation and experiment more efficient partition algorithms in the future.
+
+## FX compatibility
+
+Pipeline parallelism auto partition capability needs more advanced knowledge in order to accurately divide a model into a balanced partition set.  In order to achieve auto partitioning with the most balanced approach, we need to get the model execution graph and try to split the model base on the graph. Using torch.fx can give us a helpful graph representation in order for us to do more advanced partition with extensive analysis with tracing and profiling.
```
thomasw21, line 85: `torch.fx` gives a representation of a function no? Typically in case of a module we'd trace `forward`. However a lot of model have "secondary" forward function, typically `infer` which might be slightly different from the forward function. How does optimizing on the `forward` function, generalises to the `infer` function.

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+in order to apply the algorithm, we need to assign each top-level submodule with a proper cost, we calculate the cost based on memory and execution costs.
+
+
+### Cost calculation
+
+The cost of a submodule is defined as the following:
+
+```
+cost(m) = time_weight * execution_time(m) 
+            + memory_weight * memory_cost(m)
```
thomasw21, line 171: Why would memory weight be used? Isn't it binary as in "does it fit in memory"?

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,255 @@
...
+## Potential issues
+
+**What if a model with a big memory requirement that couldn’t fit into a single GPU?**
+A: we construct the model on CPU and do fx tracing on CPU, after the partition we move it to the corresponding device
+
+**What if a “module” (submodule) with a big memory requirement that couldn’t fit into a single GPU?**
+Can we use ShardedTensor to shard the module parameters? or we could do operation level tracing, partition this module into 2 submodules, then assign to different devices. 
+
+**What if constructing the model on CPU itself is hard?** 
+User created the model first on meta device, we use the model that’s not materializing to do symbolic tracing, but we couldn’t do profiling since it’s not materialized yet, after we do symbolic tracing, we should use a simple partition algorithm (i.e. only count the param sizes) to do the partition, then materialize afterwards.
```
thomasw21, line 243: Is there a documentation on "meta" device? I don't seem to find any.

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
deepakn94: Yes, the idea would be to give the partitioning function all devices that you have available (let's say 128 GPUs), and the system automatically figures out how to split GPUs between data and pipeline parallelism (and perhaps tensor parallelism at some point as well).

RFC-0020-Distributed-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,209 @@
...
+
+This is an RFC for the strategic plan for further developing pipeline parallelism in PyTorch. **We invite our users and partners to comment on this plan and the corresponding technical plan** to help us develop the best APIs for PP in PyTorch.
+
+Goal: *Provide a flexible, composable, and reconfigurable interface for pipeline parallelism in PyTorch that allows scaling a wide variety of PyTorch models on a wide variety of hardware configurations*.
+
+## Motivation
+
+* Pipeline parallelism (PP) is used as a lower-communication-volume technique for model parallelism. It is especially applicable when data must be transmitted across comparatively slower interconnects.
+* Several research-oriented frameworks exist that implement PP (e.g. fairscale, megatron, deepspeed), but we would like to provide a production-quality implementation and support contract for PP.
+* The existing PP implementation in PyTorch (`torch.distributed.pipeline.sync`) only supports intra-host pipeline parallelism across GPUs and does not support techniques like 1F1B scheduling. We can deliver inter-host pipelining and other features.
```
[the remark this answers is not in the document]
jamesr66a: Great, will add this. Based on all the feedback, this definitely seems like a top-line concern for the design

RFC-0020-Distributed-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,209 @@
...
+HF transformers [wants to](https://github.com/huggingface/transformers/issues/13690) incorporate 3d parallelism including Pipeline Parallelism, however the [current PyTorch implementation](https://github.com/pytorch/pytorch/blob/9f4e004abd8c5d11fc23f4ab705328cb9b4050bb/torch/distributed/pipeline/sync/pipe.py#L220) has limitations that we should address (Px is a priority, with lower x being higher priority. We assigned these priorities based on a) user need and b) implementation time/complexity, but we can adjust them based on user feedback):
+
+* Frontend limitations:
+    * **P0**: Cannot pass arbitrary data types between pipeline stages
+    * **P0**: Unclear composability in 3d parallelism scheme (data, pipeline, model parallel)
+    * **P1**: User needs to rewrite their model as an `nn.Sequential` instance
+* Backend Limitations:
+    * **P(-1)**: No cross-host support for PT pipeline parallelism API
+    * **P0**: No support off-the-shelf schedules (1F1B or interleaving)
+    * **P1**: No support arbitrary programmable schedules
```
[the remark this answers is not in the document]
jamesr66a: How about we push the `nn.Sequential` requirement up to P0 instead? We're running the risk of having too many top-line priorities that way, but we can see how it goes

RFC-0020-Distributed-Pipeline-Parallelism.md
jamesr66a: `local` means within a node, i.e. it's drawing a distinction between `torchgpipe`, which has a single driver process issuing commands onto CUDA streams v.s. something that operates across multiple nodes (and which necessarily will need to have multiple processes)

RFC-0020-Distributed-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,209 @@
...
+* Non-requirements:
+    * Composability with ZeRO-2/3 is not required. Theoretically possible, but reportedly will not give any perf gain.
+* Success Criteria:
+    * **to be determined**: Feedback on this would be appreciated
```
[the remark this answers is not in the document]
stas00: My apologies. I meant Deepspeed ZeRO - which requires no mods to the model (most of the time) - I edited my comment.

Deepspeed PP is not user friendly at all compared to that as it too wants a sequential model.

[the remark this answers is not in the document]
stas00: The only reason not to use ZeRO to scale the model - is when interconnects are slow. When interconnects are fast there is no real need for PP+TP, ZeRO-DP and PP+TP+DP perform on par.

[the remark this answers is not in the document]
stas00: This is why we ended up using Megatron-Deepspeed for the BigScience project. The network on JeanZay is slow, so PP was the only way to keep TFLOPs high.

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
```diff
@@ -0,0 +1,620 @@
...
+
+### Expanding Predication to Include the Training Loop
+
+The next logical step is to expand this proposal to encompass the training loop, allowing the user to continue to write their arbitrary Python training loop, but overlaying pipeline parallelism semantics on this using predicated execution.
+
+We can pull out the parts of the canonical training loop from the beginning and investigate how they should be executed under a predicated, pipeline parallel training loop:
+
+**Data Loader**
+
+The data loader should only load input data on rank 0. We can view this as predicating `true` on rank 0 and false on all other ranks. We can also commingle this with `recv` for stages != 0. i.e. under the hood the dataloader object will return an input micro-batch on rank 0, but will return the intermediate value received from `rank - 1` on all ranks != 0.
```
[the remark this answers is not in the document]
jamesr66a: @thomasw21 if this is something that is an expected use case, we should discuss this further as it has broader implications on the rest of the RFC and we should likely re-do all of the designs.

To help me understand: what would the execution pattern look like when pipelining a non-sequential (arbitrary DAG) model? In particular, can there exist disconnected nodes  which must be scheduled somehow? For example:

```
def forward(self, x):
  y = torch.zeros(3, 4) # On forward, must be scheduled for execution independently of the function inputs
  z = self.layer_1(x) # rank 0
  a = self.layer_2(z + y) # rank 1
  return a
```

(Hypothetically we could also have some sort of sink node in the forward that results in a disconnected edge in the autograd graph, but I can't think of a concrete scenario where that would happen)

RFC-0022-Model-Partitioning-in-Pipeline-Parallelism.md
byronyi: https://github.com/pytorch/rfcs/blob/rfc-0005/RFC-0005-structured-kernel-definitions.md

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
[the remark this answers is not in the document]
thomasw21: I would expect that disconnected graphs should work. Typically when using absolute positional embeddings, we obtain the position embeddings like the following:

```
def forward(self, x):
   pos_emb = self.get_position_embeddings() # returns nn.Parameter
   token_emb = self.compute_token_embeddings(x)
   return pos_emb + token_emb
```

However now that I think about it, it's highly unlikely that there exist a rank that holds a parentless node (otherwise you would just fuse it with one of its children, the first one in the rank ordering, and use that as a proxy parent):


```
pos_emb -------> output
             /
token_emb --
```

to

```
token_emb ----> pos_emb + token_emb ----> output
```

The reason I think of DAG, is because let's say you have ensembling techniques, where you have 5 experts, which are then aggregated. I'd expect the 5 experts to be able to run in parallel. Now imagine each expert is held inside one rank each. You'd have something like 6 ranks (5 + 1 to run experts and then aggregate).

Current framework would work like this in my understanding:
```
expert 0 ----> expert 1 + expert 0 ----> expert 2 + expert 1 + expert 0 ---> etc 
```

Each rank would actually hold its own expert, run that expert, but actually keep forwarding previous expert results to next rank no? This means that fundamentally we should only be able to run that in a sequential fashion? I would have expected all experts to be able to run in parallel. The agregator can still require to receive values from its experts in an arbitrary order (but that means we're bottlenecked by the slowest expert, instead of the sum all experts).

Does that make sense?

deepakn94:
Oh, another thing that I don't see mentioned anywhere: it is common for language models to have layers that "share" parameters. For example, the embedding layer at the front of the model and the LM head often share weights. There are a couple of different ways of handling this: a) ensure that all layers sharing parameters are on the same device, and let autograd do the right thing, b) allow layers sharing parameters to reside on different devices, but then synchronize gradients at the end of a step (and also initialize parameters the same way on all devices with the same "shared" parameters.

I am sure there are other ways of handling this as well.

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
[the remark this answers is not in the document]
jamesr66a: @thomasw21 I think modeling arbitrary graph programs in a pipeline parallel API is certainly possible (e.g. it was done in the `fairscale` experimental distributed pipeline prototype https://github.com/facebookresearch/fairscale/blob/7c2c3e0046eb3a1b1c98fd24f61fb09f90f6817f/fairscale/experimental/nn/distributed_pipeline/graph.py#L56).

OTOH, I would ask whether we can decompose this problem into simpler solutions that can be composed with each other. For example, in the first example, could we group the `get_position_embeddings()` call and the `compute_token_embeddings` call into a single pipeline stage and use RPCs to run those in parallel on remote hosts. Similarly, for an MoE architecture, can we commingle each expert into a single pipeline stage, but within that pipeline stage use remote RPCs to run the experts on remote hosts in parallel?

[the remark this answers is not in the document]
stas00: Won't that be incredible inefficient speed-wise comparatively to other stages with a lot more logic in them? Also wrt to memory usage - as you will want to partition each stage so that memory usage will be balanced.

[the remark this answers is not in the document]
jamesr66a: @stas00 could you clarify? Are you talking about the examples specifically or the general concept of calling RPCs within a pipeline stage?

In terms of latency, it should be equivalent to the critical path through the (2 or 5) parallel calls plus some RPC overhead. If these things are actually running on the same box, you could also just issue those two commands on different CUDA streams and get parallelism that way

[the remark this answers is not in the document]
stas00: I'm thinking in the terms of pipeline stages as they are currently used by Megatron and Deepspeed.

A typical pipeline for a transformer block is just one or more transformer blocks with an appendix of the embedding in the first pipeline stage, e.g.:

```
[emb|trans|trans] -> [trans|trans] -> [trans|trans|embed]
```
which ensures that they are mostly of the same size and do about the same amount of processing.

but if we now introduce a much finer granularity stages such as a stage that computes `get_position_embeddings` which is super quick and requires almost no memory I'm no longer sure what happens.

Unless your definition of a pipeline stage is something else.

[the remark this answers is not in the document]
jamesr66a: @stas00 agreed that splitting those embeddings functions off into stages is dubious, but I was going off of @thomasw21's original formulation, so I would extend that question to him.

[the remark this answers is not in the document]
thomasw21: Yeah sorry, when mentioning embeddings I was trying to think of examples of parentless nodes. I think that it's highly unlikely that you would create a pipeline stage that doesn't require a previous stage to be ran before or depend on any inputs.

I think the part I'm perhaps a bit unfamiliar is the concept of having multiple hosts on a single pipeline stage. I'm mostly familiar with deepspeed's version. Stages are assigned a single device. We mainly get parallel execution thanks to microbatching/scheduling.

Taking the multi experts again. If I understand correctly there'd be two stages, one to run all the experts, and another one to run the aggregator? If so that makes sense to me. The rules should become like:
 - Within a single pipeline stage, multiple hosts run in parallel.
 - Ranks feed their results sequentially to next ranks.
 
But then if you manually have to setup RPCs, aren't you manually doing Pipeline parallelism? ie the whole model is a single stage and use RPCs to schedule calls on different devices?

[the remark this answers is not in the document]
jamesr66a: @thomasw21 here's what i'm describing, maybe you have something else in mind:

Given the definitions of pipelining used in the literature, pipeline parallelism looks like turning the following:

```
def my_model(x):
  x = op1(x)
  x = op2(x)
  x = op3(x)
  return x
```

Into something like (suppose we cut between op2 and op3 into 2 stages):

```
def my_model0(x):
  x = op1(x)
  x = op2(x)
  return x

def my_model1(x):
  x = op3(x)
  return x
```

And then the composition `my_model1(my_model0(x))` would run in pipelined execution. I think in the MoE example, the original source would look something like:

```
def my_model(x):
  x = op1(x)
  x = op2(x)
  x = op3(x)

  gates = gate_fn(x)
  moe_results = []
  for i, expert in enumerate(experts):
    moe_results.append(expert(x[i]))
  x = combine(moe_experts)
  return x
```

Suppose we apply the asme pipelining policy, i.e.:

```
def my_model0(x):
  x = op1(x)
  x = op2(x)
  return x

def my_model1(x):
  x = op3(x)

  gates = gate_fn(x)
  moe_results = []
  for i, expert in enumerate(experts):
    moe_results.append(expert(x[i]))
  x = combine(moe_results)
  return x
```

As you point out, there is inherent parallelism in the parallel application of the experts, i.e. in the `enumerate(experts)` loop. However, with respect to how the code appears on the screen, this is "horizontal" parallelism vs. the "vertical" parallelism of pipelining. So my proposal is to basically implement horizontal parallelism in a way that is orthogonal to pipelining, i.e.:

```
def my_model0(x):
  x = op1(x)
  x = op2(x)
  return x

def my_model1(x):
  x = op3(x)

  gates = gate_fn(x)
  futures = []
  for i, expert in enumerate(experts):
    futures.append(rpc(expert(x[i])))

  moe_results = []
  for fut in futures:
    moe_results = fut.wait()

  x = combine(moe_results)
  return x
```

In this way, horizontal parallelism is achieved while keeping the pipeline parallel system simple and focused on vertical parallelism. The same argument can be made for model parallelism, i.e. pipeline parallelism can operate on ShardedTensor instances that dispatch to collectives, but pipeline parallelism itself does not need to be aware of this scheme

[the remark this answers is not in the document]
jamesr66a: @thomasw21 Put another way, pipeline parallelism is about parallelizing code _across micro-batches_, whereas the MoE parallelism is _within a micro-batch_, thus I think these should be orthogonal to each other. Does that make sense?

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
```diff
@@ -0,0 +1,651 @@
...
+* (**D5**) `nn.Sequential` not needed, but potentially an `nn.Graph` instance may be needed in some cases
+* (**D6**) Async is probably supportable but not clear. From their presentation, the actor/register model with backpressure can implement on-demand data loading, but I’m not 100% sure what that API looks like
+* (**D7**) Unified programming model that already exists
+
+**Con**
+
+* (split pro/con) (**D2**) Not clear if `1f1b` or other schedules are implemented/implementable?
+* (**D8**) Not clear what the training loop abstraction looks like. The optimizer is installed via an `nn.Graph` API. Loss calculation is created in the `nn.Graph.build()` method.
+
+## Approach 7: Varuna
```
jamesr66a, line 548: @stas00 I've added details about Varuna in this section, please have a look and let me know if I've missed anything

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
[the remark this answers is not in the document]
thomasw21: I see, this makes sense. Would the partitioning algorithm be able to work orthogonally with having these kind of systems but in place?

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
```diff
@@ -0,0 +1,672 @@
...
+
+The workflow of Varuna looks like the following:
+* The user manually annotates their model with [CutPoints](https://github.com/microsoft/varuna/blob/ea13ce6a8934bfe829b662a56af4dc29044fcb34/docs/cutpoint.rst). These are points in the program where the system _may_ place a pipeline stage
+* The user wraps their model in the [Varuna](https://github.com/microsoft/varuna/blob/ea13ce6a8934bfe829b662a56af4dc29044fcb34/docs/varuna.rst#id9) class and configures pipeline parallelism via this interface. This includes parameters like chunk size, installing the optimizer, and informing the system of the rank of the current running instance.
+    * Internally, the system is going to do a roudn of profiling to determine the optimal pipeline balance and choose a subset of the cut-points together to represent the code for different pipeline stages.
```
stas00, line 571:
```suggestion
    * Internally, the system is going to do a round of profiling to determine the optimal pipeline balance and choose a subset of the cut-points together to represent the code for different pipeline stages.
```

stas00 (commented):
Thank you for adding Varuna, @jamesr66a!

The notes look good

SeanNaren:
This is a fantastic summary of attempts for PP, great work @jamesr66a :)

Within Lightning we experimented with the initial `RemoteModule` from Fairscale which relied on RPC, however as discussed the requirement of everything being sequential made it very restrictive.

There is another example (albeit very intrusive in the user's code) by the GraphCore team where PP is a requirement for many models when using IPUs https://docs.graphcore.ai/projects/poptorch-user-guide/en/1.0.0/overview.html#poptorch-block-and-poptorch-beginblock

Even though this approach is extremely intrusive, it does have its merits of support a wider range of models + being a bit more expressive. I would relate this to the idea behind [FlexFlow](https://github.com/flexflow/FlexFlow)'s torch.fx support as we would need to traverse the graph.

stas00:
>Even though this approach is extremely intrusive, it does have its merits of support a wider range of models + being a bit more expressive. I would relate this to the idea behind FlexFlow's torch.fx support as we would need to traverse the graph.

Indeed, I was considering mentioning https://github.com/flexflow/flexflow as another framework to consider but last we checked it still had the PP-support only planned.

RFC-0020-Distributed-Pipeline-Parallelism.md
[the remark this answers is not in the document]
tchaton: Hey @jamesr66a,

I agree with @stas00 on this one. I believe adding PP support for more models would definitely be a criterion of success and unblock an integration within PyTorch Lightning as we used to support the experimental Pipe version from FairScale, but it was too limited and unstable.

RFC-0020-Distributed-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,211 @@
...
+This section is meant to capture key users/researchers who would benefit from such a pipeline parallelism API. **We invite additional comments to fill in users/researchers who would benefit from this API and would like to see their requirements satisfied**.
+
+#### P0: HF Transformers
+
+HF transformers [wants to](https://github.com/huggingface/transformers/issues/13690) incorporate 3d parallelism including Pipeline Parallelism, however the [current PyTorch implementation](https://github.com/pytorch/pytorch/blob/9f4e004abd8c5d11fc23f4ab705328cb9b4050bb/torch/distributed/pipeline/sync/pipe.py#L220) has limitations that we should address (Px is a priority, with lower x being higher priority. We assigned these priorities based on a) user need and b) implementation time/complexity, but we can adjust them based on user feedback):
+
+* Frontend limitations:
+    * **P0**: Cannot pass arbitrary data types between pipeline stages
+    * **P0**: Unclear composability in 3d parallelism scheme (data, pipeline, model parallel)
+    * **P1**: User needs to rewrite their model as an `nn.Sequential` instance
```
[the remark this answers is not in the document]
jamesr66a: This has been updated elsewhere to be a P0, let me fix up this location

RFC-0020-Distributed-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,211 @@
+# [RFC] Ceci n'est pas pipeline parallelism (Pipeline Parallelism 2021Q4/2022 Plan)
```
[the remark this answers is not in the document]
jamesr66a: a small joke :) https://en.wikipedia.org/wiki/The_Treachery_of_Images

rohan-varma:
cc @pritamdamania87 @pbelevich @mrshenli @zhaojuanmao

RFC-0021-Distributed-Pipeline-Parallel-Technical.md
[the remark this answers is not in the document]
stas00: To support the discussion of using MoE in the pipeline

Microsoft and Google have just published architectures that use MoE:

https://ai.googleblog.com/2021/12/more-efficient-in-context-learning-with.html
https://www.deepspeed.ai/news/2021/12/09/deepspeed-moe-nlg.html

Deepspeed's version uses Megatron-Deepspeed so it's already based on Deepspeed PP implementation - so you could probably check how they dealt with MoE - the Moe Branch is here: https://github.com/microsoft/Megatron-DeepSpeed/tree/moe-training

But reading closer Deepspeed's version uses only top-1 expert, so it's then irrelevant to this discussion, as it runs only 1 MoE.

Google's arch uses 2 MoEs and thus is relevant as it needs to parallelize the processing of these 2.

RFC-0020-Distributed-Pipeline-Parallelism.md
```diff
@@ -0,0 +1,209 @@
...
+### P(-1): Implement cross-host support for pipeline parallelism
+
+Existing approaches that support this (in no particular order):
+
+* Fairscale [experimental distributed pipeline parallelism](https://github.com/facebookresearch/fairscale/tree/main/fairscale/experimental/nn/distributed_pipeline)
+* Sagemaker [model parallelism](https://arxiv.org/abs/2111.05972)
+* [DeepSpeed pipeline parallelism](https://www.deepspeed.ai/tutorials/pipeline/)
+* [OneFlow](https://github.com/Oneflow-Inc/oneflow)
+
+Proposed approach short-list: (all approaches can be seen in [[RFC] Distributed Pipeline Parallel Training Technical Approach](https://github.com/pytorch/rfcs/blob/master/RFC-0021-Distributed-Pipeline-Parallel-Technical.md)
```
hjm-aws, line 68: Sorry, the link is now broken.

PR #34, opened 2021-12-15
RFC-0022: Proposal for adding new restart policy to torch.distributed…

aivanou:


ASDen:
First, moving that direction is a great step! thanks!

Second, I would highly suggest looking into [Elastic Horovod](https://horovod.readthedocs.io/en/stable/elastic_include.html), I think they managed to get few things right that would be big additions here:
- Graceful node removal = continuing as-is without losing a single batch!
- Providing very helpful classes for handling full state syncing (including weights, sampler, optimizer, primitive types, etc..), this is vital for simple true graceful node removal
- The idea of doing memory (not disk) checkpointing of model weights seems interesting (much cheaper), esp. given this should be a fault-tolerant distributed system

Also, would this RFC support dynamically adding new nodes while training ?

PR #35, opened 2022-01-08
Update the Backward Compatibility Service Level Agreement (SLA)

iseeyuan:
* __->__ #35

We are reaching to a consensus to an SLA window (6 months) for both mobile and server, both internal and OSS. Update it accordingly in this RFC.

### DIFF
--- a/RFC-0017-PyTorch-Operator-Versioning.md
+++ b/RFC-0017-PyTorch-Operator-Versioning.md
@@ -7,9 +7,9 @@ These operators sometimes require changes to maintain the high quality user expe
 
 BC and FC breaking changes have been challenging to coordinate across PyTorch because there are multiple consumers of PyTorch’s op set and we promise to keep models running in production working as expected.
 
-We are providing the same Service Level Agreement (SLA) to both internal and external use cases.
-
 This document proposes a new BC and FC policy based on operator versioning.
+
+Moving forward, we're not having a difference between Meta internal and Open Source (OSS) guarantees. They would be moving under *the same Service Level Agreement (SLA)* to both internal and external use cases.
 
 
 ## History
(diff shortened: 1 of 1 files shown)


RFC-0017-PyTorch-Operator-Versioning.md
```diff
@@ -19,12 +19,11 @@ This document proposes a new BC and FC policy based on operator versioning.
...
 Backwards compatibility (BC), the ability for PyTorch to continue running programs from older versions, is important so programs don’t need to be forcefully updated to comply with the new runtime implementation.
 
-PyTorch current SLA on backwards compatibility:
+PyTorch SLA on backwards compatibility:
 
 
 
-* **OSS** — “stable” features will be deprecated for one release before a BC-breaking change is made. [PyTorch OSS BC-breaking policy](https://pytorch.org/docs/master/) 
-* **FB Internal** — we will not break a serialized torchscript program running in production at Facebook (to be replaced with a more generic SLA)
+* PyTorch SLA will ensure that models developed using a cerntain version will be supported for *six months* from the version release date.
```
cccclai, line 26: Is 6 months better or 180 days?

raziel (changes requested):
Shouldn't we have a sub-section at the end of the proposal section that describes the new SLA? Rather than modifying the "History" section

iseeyuan:
> Shouldn't we have a sub-section at the end of the proposal section that describes the new SLA? Rather than modifying the "History" section

Good catch! Let me move to the proposal section and update the wording.

mruberry (approved):
LGTM!

raziel:
Thanks @iseeyuan 

Should we also call out as a goal/outcome that moving forward we're not having a difference between internal and OSS guarantees since they would be moving under the same SLA?

Also, update the Facebook / FB references :-)

raziel (approved):
Looks good, aside from the point I raised on internal vs oss

iseeyuan:
> Thanks @iseeyuan
> 
> Should we also call out as a goal/outcome that moving forward we're not having a difference between internal and OSS guarantees since they would be moving under the same SLA?
> 
> Also, update the Facebook / FB references :-)

Sounds good! Will call out the outcome, and update "Facebook". Thanks!

PR #39, opened 2022-02-01
RFC-0025 Improving incremental builds

peterbell10:
This RFC proposes changes to ATen that will allow more granular header dependencies and tools to enforce their usage across the codebase which should greatly improve incremental and cached build performance.

Note that a lot of this has already been implemented and merged but this RFC should provide a complete picture of the motivation and how these PRs fit together. 
- The stack starting with https://github.com/pytorch/pytorch/pull/63612 introduces `TensorBase` and makes various CUDA files `TORCH_ASSERT_NO_OPERATORS` compliant, as does https://github.com/pytorch/pytorch/pull/66620.
- The stack starting with https://github.com/pytorch/pytorch/pull/68246 generates the per-operator headers and makes various code-generated files `TORCH_ASSERT_ONLY_METHOD_OPERATORS` compliant.
- The unmerged stack starting with https://github.com/pytorch/pytorch/pull/68462 converts most of ATen to use per-operator headers.

### DIFF
--- a/RFC-0025-improving-incremental-builds.md
+++ b/RFC-0025-improving-incremental-builds.md
@@ -1,0 +1,217 @@
+# Improving incremental builds
+**Authors:**
+@peterbell10
+
+## **Summary**
+
+This RFC proposes changes to ATen that will allow more granular header
+dependencies and tools to enforce their usage across the codebase.
+This should greatly improve incremental and cached build performance.
+
+## **Motivation**
+
+In ATen, certain generated headers files include the definitions of
+every operator in PyTorch. Consequently any new operator, or update
+to any of ATen's over 1000 existing operators in
+`native_functions.yaml`, requires a full rebuild of ATen and torch.
+These problematic headers include `Tensor.h`, `Functions.h`,
+`NativeFunctions.h`, which are all widely used and currently
+unavoidable.
+
+For example, adding a new operator that isn't referenced anywhere
+results in 1000+ files being rebuilt over many hours of CPU time
+(single threaded). After the proposed changes you can expect at most a
+small handful of files to be rebuilt, most of which should contain
+code generated specifically for that new operator.
+
+## **Proposed Implementation**
+
+This RFC proposes two changes to allow breaking up these header
+dependencies, and avoid unnecessary header dependencies in ATen.
+
+1. Add a base class to `Tensor` called `TensorBase` which has no
+   method operators but is otherwise fully interoperable. This is
+   particularly useful for kernel code which often doesn't call any
+   operators but still needs to pass `Tensor`s around.
+2. Split code generated header files up on a per-operator basis, so
+   only necessary operators need to be included. These headers still
+   depend on `Tensor.h` so will depend on method operators.
+
+Additionally, two macros are introduced which can be defined to
+enforce strict adoption of these more granular headers:
+
+- `TORCH_ASSERT_NO_OPERATORS` when defined, will make inclusion of any
+  code generated from the operator list into a compiler error. In
+  practice this limits you to only use `TensorBase` and `TensorIterator`.
+- `TORCH_ASSERT_ONLY_METHOD_OPERATORS` which allows `Tensor.h` to be
+  included, but not the other umbrella headers. Instead specific
+  operators must be included directly as they are used.
+
+These will be implemented by inserting preprocessor guards into the
+relevant header files, with `ATen/core/TensorBody.h` as an example
+```cpp
+#ifdef TORCH_ASSERT_NO_OPERATORS
+#error This change adds a dependency on native_functions.yaml,            \
+  meaning the file will need to be re-compiled every time an operator     \
+  is changed or added. Consider if your change would be better placed in  \
+  another file, or if a more specific header might achieve the same goal. \
+  See NOTE: [Tensor vs. TensorBase]
+#endif
+```
+
+This assists in first detecting problematic includes, even when
+buried in layers of transitive includes. Then further with
+maintaining the header hygiene into the future, as developers get
+immediate feedback when undesirable headers are added in the future.
+
+### TensorBase
+`Tensor` serves two related but separate functions. It is both a
+user-friendly handle to the underlying `TensorImpl` data structure,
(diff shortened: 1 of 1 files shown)


RFC-0025-improving-incremental-builds.md
```diff
@@ -0,0 +1,217 @@
...
+compliance often requires significantly restructuring the code.
+However, where the cost-benefit is favorable, this macro should be
+defined manually in the source file. This will include files that
+already use `TensorIterator` exclusively and just need their includes
+pruned; and more costly to compile files where taking time to
+restructure them is worthwhile.
+
+## **Metrics**
+
+Some key metrics that should be improved by these changes:
```
albanD, line 169: Do we have some measure of these metrics. Or a snapshot of todays values so that we can measure future improvements.

RFC-0025-improving-incremental-builds.md
```diff
@@ -0,0 +1,217 @@
...
+2-3 hours of (serial) CPU time.
+
+
+## **Prior Art**
+Not aware of any relevant prior art.
+
+## **How we teach this**
+This change doesn't effect end-users.
+
+For PyTorch developers, documentation should be written on how to
```
albanD, line 205: Could you extend on this one? Currently this is mainly documented in the compile time error messages. Should we have a wiki entry about why and how to do the right thing?

RFC-0025-improving-incremental-builds.md
```diff
@@ -0,0 +1,217 @@
...
+write code in compliance with the two enforcement macros and linked in
+the corresponding error messages.
+
+## **Unresolved questions**
```
albanD, line 209: Given that most of the above is done today. Do we want to extend this rfc to try and answer such questions?

RFC-0025-improving-incremental-builds.md
```diff
@@ -0,0 +1,217 @@
...
+- Is it possible/desirable to enforce `TORCH_ASSERT_NO_OPERATORS` automatically?
+  e.g. for all `.cu` files, or all files over a certain compile time.
+- Can `include-what-you-use` completely automate operator includes?
```
albanD, line 212: I could imagine a daily/weekly automatic CI that automatically runs/fix this similar to how we do for not important python lint with https://github.com/pytorch/pytorch/commit/1edf6f56477ea317af845a4cd65eb311737961f0

Do you think that could work? Any danger of BC-breaking change for c++ users?

RFC-0025-improving-incremental-builds.md
```diff
@@ -0,0 +1,217 @@
...
+restructure them is worthwhile.
+
+## **Metrics**
+
+Some key metrics that should be improved by these changes:
+- Time to rebuild after adding editing a functional operator
+- Time to rebuild after adding editing a method operator
+- `sccache` miss rate in open source CI
+
+## **Drawbacks**
```
albanD, line 174: Some of our users include internal file and rely on the fact that most of ATen is available at that point.
This series of change (by cleaning core) will break them. Is there anything we can do to reduce such breakage while still fixing our build?

RFC-0025-improving-incremental-builds.md
```diff
@@ -0,0 +1,217 @@
...
+write code in compliance with the two enforcement macros and linked in
+the corresponding error messages.
+
+## **Unresolved questions**
+- Is it possible/desirable to enforce `TORCH_ASSERT_NO_OPERATORS` automatically?
+  e.g. for all `.cu` files, or all files over a certain compile time.
+- Can `include-what-you-use` completely automate operator includes?
+  Tools exist for strictly managing _all_ includes, but that would be
+  a significant change from existing include style.
+- Is it worth adopting `TensorBase` more widely than just kernels?
```
albanD, line 215: I think we could answer this by seeing how much benefit we will get from doing that? Is there an easy way to see how this will change the metrics discussed above?

RFC-0025-improving-incremental-builds.md
peterbell10: I've added a table with some measurements. The "after" measurements are incomplete of course because I'm not finished yet and much of ATen and torch still include `ATen.h`.

RFC-0025-improving-incremental-builds.md
peterbell10: Yes, I a wiki entry makes the most sense I think.

RFC-0025-improving-incremental-builds.md
peterbell10: `iwyu` does have a "safe mode" that never remove includes from header files so I think it's possible without BC-breaking changes. The main worries are that it sometimes requires manual input like telling it which headers are public or private (e.g. `List-inl.h` is private and `List.h` is public) and also additional in-source pragmas (e.g. to allow `List.h` to include `List-inl.h`).

I think the `ATen/ops` headers are fairly safe though so it may be possible to filter the iwyu output to only change operator headers.

RFC-0025-improving-incremental-builds.md
peterbell10: I suppose we could do something like 
```
#ifndef CAFFE2_BUILD_MAIN_LIB
#include <ATen/ATen.h>
#endif
```

However, I think that user code should really just be including `ATen.h` itself.

RFC-0025-improving-incremental-builds.md
peterbell10: I don't know of an easy way but some numbers to put it into perspective: 1184 files include `Tensor.h` today which is ~70% of the compile time for ATen and torch. ~40% of operators have some method variant so keeping it this way means very roughly speaking we can expect ~60% of operator changes to have fast incremental builds.

However, I suspect that methods change less often and that more development effort will be going into new operators which are mostly functional-only. So in practice it may be much more the 60% that are fast.

RFC-0025-improving-incremental-builds.md
albanD: I think it's good to keep updating that table so that we have a place with the historical data and improvements as we move forward.

albanD:
The CUDA builds are not mentioned in this doc. Does it need any special casing?

PR #40, opened 2022-02-16
[PyTorch] Add RFC for lightweight dispatch

larryliu0820:
* __->__ #40


RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
+# Context
+With recent developments in PyTorch Edge applications on embedded systems and resource limited devices, the issue of op registration/dispatching runtime overhead and build-time complexity has risen to the fore.
+
+We thought about possible solutions:
+* One option is to keep using the torch-library C++ API to register/dispatch ops, but this will necessitate careful cost-cutting for these use cases with more and more intrusive “#ifdef” customizations in the core framework.
```
iseeyuan, line 5: Adding some background of torch-library API would help.

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+* The other option (which is what we propose here) is to utilize the function schema yaml file to "declare" ops and then use the codegen framework to generate lightweight code for runtime execution.
+
+The essential point of this proposal is that the function schema DSL (which we use to declare the standard ATen ops) combined with the codegen framework (which we use to generate the registrations, dispatch stubs and other “glue” code for ATen ops) is the bare minimum set of reusable tools for building custom extensions that are compatible with the PyTorch ecosystem.
+
+# Motivation
+* **Performance**
+  * For recent use cases of Edge interpreter, we need to satisfy more and more strict initialization latency requirements, where analysis shows op registration contributes to a large portion of it.
```
raziel, line 12: Rename edge>mobile

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+  * This project creates an opportunity to reduce binary size by getting rid of the dispatcher and enables further size optimization on unboxing wrappers.
+* **Ability to incorporate custom implementation of ATen ops**
+  * For some of the edge use cases, we need to support custom implementations of ATen ops. With an extra op registration path such as codegen unboxing it is easier to hookup ops with custom native functions.
+
+# Overview
+![codegen drawio](https://user-images.githubusercontent.com/8188269/154173938-baad9ee6-0e3c-40bb-a9d6-649137e3f3f9.png)  
+
+
+Currently the lite interpreter (or Edge runtime) registers all ATen ops into the dispatcher and some other ops into the JIT op registry. At model inference time the interpreter will look for the operator name in the JIT op registry first, if not found then it will look into the dispatcher. This proposal **adds a build flavor that moves these ATen ops from dispatcher to JIT op registry** so that it’s easier to optimize (e.g., avoid schema parsing) and can also reduce dependencies.
```
raziel, line 26:
```suggestion
Currently the mobile interpreter registers all ATen ops into the dispatcher and some other ops into the JIT op registry. At model inference time, the interpreter will look for the operator name in the JIT op registry first, if not found then it will look into the dispatcher. This proposal **adds a build flavor that moves these ATen ops from dispatcher to JIT op registry** so that it’s easier to optimize (e.g. avoid schema parsing) and reduce dependencies. 
```

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+
+The interpreter is looking for a boxed function but our native implementation is unboxed. We need “glue code” to hook up these two. This proposal **extends the capabilities of codegen to generate the unboxing wrappers for operators**, as well as the code to register them into the JIT op registry. The interpreter will call generated unboxing wrappers, inside these wrappers we pop out values from the stack, and delegate to the unboxed API.
+
+To avoid hitting the dispatcher from the unboxed API, we will choose static dispatch so that we hit native functions from the unboxed API directly. To make sure we have feature parity as the default build, this proposal **adds support for multiple backends in static dispatch**.
+
+In addition to that, this proposal also supports features critical to Edge, such as **tracing based selective build** and **runtime modularization** work.
```
raziel, line 32:
```suggestion
In addition to that, this proposal also supports features critical to mobile use cases, such as **tracing based selective build** and **runtime modularization** work.
```

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+
+We are going to break the project down into three parts, for **step 1 we are going to implement the codegen logic** and generate code based on [native_functions.yaml](https://fburl.com/code/2wkgwyoq), then we are going to verify the flow that we are able to find jit op in the registry and eventually call codegen unboxing wrapper (the red flow on the left). **Step 2 will focus on how to make sure we have feature parity** with the original op registration and dispatch system, with tasks like supporting multiple backends in static dispatch, supporting custom ops as well as custom kernels for ATen ops.For **step 3 we are going to integrate with some target hardware platforms**. These are the problems we need to address in step 3 including: avoiding schema parsing at library init time, supporting tracing based selective build. The goal of step 3 is to make sure per-op codegen unboxing works for our target hardware platforms and is ready to ship to production.
```
raziel, line 38:
```suggestion
We are going to break the project down into three parts, for **step 1 we are going to implement the codegen logic** and generate code based on [native_functions.yaml](https://fburl.com/code/2wkgwyoq), then we are going to verify the flow that we are able to find jit op in the registry and eventually call codegen unboxing wrapper (the red flow on the left). **Step 2 will focus on how to make sure we have feature parity** with the original op registration and dispatch system, with tasks like supporting multiple backends in static dispatch, supporting custom ops as well as custom kernels for ATen ops. For **step 3 we are going to integrate with some target hardware platforms** to validate latency and binary size improvements. These are the problems we need to address in step 3 including: avoiding schema parsing at library init time, supporting tracing based selective build. The goal of step 3 is to make sure per-op codegen unboxing works for our target hardware platforms and is ready to ship to production use cases.
```

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+
+
+### Step 1
+
+Bring back the unboxing kernel codegen using the new codegen framework. And make the registration no-op when we turn on the static root-op dispatch for lightweight dispatch use cases. All tasks in step 1 are based on the server version of PyTorch interpreter.
+
+
+#### Codegen core logic
+
+These tasks will generate C++ code that pops ivalues out from a stack and casts them to their corresponding C++ types. This core logic should be shared across two types of codegens so that it can be covered by all the existing tests on server side.
```
raziel, line 48:
```suggestion
These tasks will generate C++ code that pops IValues out from a stack and casts them to their corresponding C++ types. This core logic should be shared across two types of codegens so that it can be covered by all the existing tests on server side.
```

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+
+
+
+* **JIT type -> C++ type**. This is necessary for some of the optional C++ types, e.g., we need to map `int` to `int64_t` for the last argument in the example.
+    * This is already done in [types.py](https://github.com/pytorch/pytorch/blob/master/tools/codegen/api/types.py), and we need to integrate it into our new codegen.
+* **JIT type -> IValue to basic type conversion C++ code.** E.g., the first argument of this operator: `Tensor(a) self` needs to be translated to: `(std::move(peek(stack, 0, 4))).toTensor()`
+    * IValue provides APIs to directly convert an ivalue to these basic types. See [ivalue_inl.h](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/core/ivalue_inl.h#L1453-L1493)
```
raziel, line 55:
```suggestion
    * IValue provides APIs to directly convert an IValue to these basic types. See [ivalue_inl.h](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/core/ivalue_inl.h#L1453-L1493)
```

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+    * Here’s a [list](#bookmark=id.deyvpbsb5yel) of all the JIT types appearing in native_functions.yaml, most of them can be converted using ivalue’s API.
```
raziel, line 56:
```suggestion
    * Here’s a [list](#bookmark=id.deyvpbsb5yel) of all the JIT types appearing in native_functions.yaml, most of them can be converted using IValue’s API.
```

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+
+**Bringing codegen unboxing to server build is out of the scope of this project.** We evaluated the option of replacing JIT op registration hook (`register_c10_ops.cpp`) with codegen unboxing wrappers, but realized that effort needs proper design and a lot of effort and only brings in small value:
+
+
+
+* Having two op registration mechanisms brings more confusion.
+* For the scenario of adding a new operator (not to `native_functions.yaml`), we need to provide clear guidance to add it to the JIT op registry as well, otherwise JIT execution will break.
+* We can add tests on the mobile build for the sake of coverage.
+
+For OSS mobile integration, we will need to have a new build flavor to switch between c10 dispatcher vs jit op registry. This new flavor will include codegen source files (`CodegenFunctions.h, CodegenFunctions.cpp, CodegenUnboxingWrappers.cpp`) instead of existing dispatcher related source files: `Operators.cpp`, `RegisterSchema.cpp `etc, similar to the internal build configuration.
```
raziel, line 131: let's clarify here again that this new path will be opt in

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+
+1. Performance gain of using JIT op registry is insignificant or even worse than dispatcher.
+    1. De-risked: from the prototype running on a target platform it is proved to save latency on initial load.
+2. Binary size regression. Need to make sure selective build works.
+3. Mobile use case requires features only available on dispatcher.
+    1. E.g., boxed fallback mechanism for [conj](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/MathBitsFallback.h#L78) operator.
+
+## Testing & Tooling Plan
+
+Expand existing tests on lite interpreter to cover codegen logic. Let the cpp test target depending on codegen unboxing library, test if a module forward result from JIT execution equals to the lite interpreter execution. Since JIT execution goes through metaprogramming unboxing and lite interpreter execution goes through codegen unboxing, we can make sure the correctness of codegen unboxing. Example:
```
raziel, line 262:
```suggestion
Expand existing tests on the mobile interpreter to cover codegen logic. Let the cpp test target depending on codegen unboxing library, test if a module forward result from JIT execution equals to the mobile interpreter execution. Since JIT execution goes through metaprogramming unboxing and mobile interpreter execution goes through codegen unboxing, we can make sure the correctness of codegen unboxing. Example:
```

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+
+ auto resd = res.toTensor();
+ auto refd = ref.toTensor();
+ ASSERT_TRUE(resd.equal(refd));
+}
+```
+
+## Appendix
+
+See design doc: [Lightweight operator registration & dispatching](https://docs.google.com/document/d/1XgJDhm0crrBNMRAwm5XXVnGvB7Z73qgsAzJTkt4FjAg/edit)
```
raziel, line 291: is this public?

raziel (commented):
Thanks!

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,287 @@
...
+=====================
+- func: acosh(Tensor self) -> Tensor
+  variants: function, method
+  structured_delegate: acosh.out
+
+- func: acosh.out(Tensor self, *, Tensor(a!) out) -> Tensor(a!)
+  structured: True
+  structured_inherits: TensorIteratorBase
+  dispatch:
+    CPU, CUDA: acosh_out
```
iseeyuan, line 160: QuantizedCPU is not listed as a dispatch (it's listed in relu_). Would code-gen consider the dispath listed here? If so, we don't need to generate QuantizedCPU for this op.

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,287 @@
...
+
+```cpp
+Functions.h
+===========
+// aten::acosh(Tensor self) -> Tensor
+TORCH_API inline at::Tensor acosh(const at::Tensor & self) {
+    DispatchKeySet _dk_set = c10::detail::multi_dispatch_key_set(tensor);
+    DispatchKey _dk = _dk_set.highestPriorityBackendTypeId();
+    switch (_dk) {
+    case DispatchKey::CPU:
```
iseeyuan, line 173: If it's runtime-control flow, does it mean that we need to build both the kernels, even if we just use one? We could use tracing based selective build, but tracing would still be required.

RFC-0020-Lightweight-Dispatch.md
larryliu0820: Right this may be a bad example, but on high level we will look at this dispatch keys and generate switch cases.

RFC-0020-Lightweight-Dispatch.md
larryliu0820: We should trace backend AOT to avoid building both kernels. I think Dhruv has a proposal of doing so.

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,287 @@
...
+* **JIT type -> IValue to TensorOptions type conversion C++ code.**
+    * Handle TensorOptions (that is not 1-1 mapping across two types of arguments), we can refer to [python.py](https://github.com/pytorch/pytorch/blob/master/tools/codegen/api/python.py#L999-L1068), maybe follow the logic over there.
+* **JIT schema -> unboxed function**. With all the arguments being translated, generate the C++ code to call the correct unboxed function and return the result (push it back to stack).
+    * Figure out how to map schema to unboxed C++ function. Reference [python.py](https://github.com/pytorch/pytorch/blob/master/tools/codegen/api/python.py#L955)
+    * Deal with method and function separately, also handle the `out` cases.
+
+
+#### Codegen source file details
+
+With the logic from the previous section, we should be able to wrap the code into a function pointer and register it into [torch::jit::OperatorRegistry](https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/runtime/operator.cpp#L19).
```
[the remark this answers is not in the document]
larryliu0820: I'm surprised it's that small. The file looks huge to me.

RFC-0020-Lightweight-Dispatch.md
```diff
@@ -0,0 +1,291 @@
...
+* The other option (which is what we propose here) is to utilize the function schema yaml file to "declare" ops and then use the codegen framework to generate lightweight code for runtime execution.
+
+The essential point of this proposal is that the function schema DSL (which we use to declare the standard ATen ops) combined with the codegen framework (which we use to generate the registrations, dispatch stubs and other “glue” code for ATen ops) is the bare minimum set of reusable tools for building custom extensions that are compatible with the PyTorch ecosystem.
+
+# Motivation
+* **Performance**
+  * For recent use cases of Edge interpreter, we need to satisfy more and more strict initialization latency requirements, where analysis shows op registration contributes to a large portion of it.
+  * With existing meta-programming based unboxing logic shared between mobile and server, it’s relatively inflexible to introduce optimizations.
+  * Also with static dispatch, we don’t have to register all of the ops into the JIT op registry, which saves runtime memory usage and further reduces static initialization time.
```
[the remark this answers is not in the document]
larryliu0820: > we don’t have to register all of the ops into the JIT op registry

I meant we only need to register the root ops.

raziel (approved):
lgtm I think there's a couple of things to follow up on, but maybe better to just get this version in.

PR #43, opened 2022-06-22
RFC-0026-logging-system

kurtamohler:
RFC for a consistent C++/Python message logging system in PyTorch

Feature was requested in https://github.com/pytorch/pytorch/issues/72948


RFC-0026-logging-system.md
```diff
@@ -0,0 +1,103 @@
...
+  - **Warning**: Emits a message as a warning. By default, this will turn into
+    a Python warning
+
+  - **Error**: Emits a message as an error. By default, this will turn into
+    a Python error
+
+  - TODO: Should we also have a **Fatal** severity for integration with
+    Meta's internal logging system? A fatal message terminates the program
+
+* Offer different classes of messages, including at least the following:
```
albanD, line 26: Should we include all of the classic python error/warnings here (TypeError, ValueError, NotImplementedError, DeprecationWarning, etc) ?
Note that we already use all of these internally

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,103 @@
...
+    [PyTorch feature classifications](https://pytorch.org/blog/pytorch-feature-classification-changes/)
+
+  - **Prototype**: Emitted when a prototype feature is called. See
+    [PyTorch feature classifications](https://pytorch.org/blog/pytorch-feature-classification-changes/)
+
+* Continue using warning/error APIs that currently exist in PyTorch wherever
+  possible. For instance, `TORCH_CHECK`, `TORCH_WARN`, and `TORCH_WARN_ONCE`
+  should continue to be used in C++
+
+  - NOTE: These existing APIs don't currently have a concept of message classes,
```
albanD, line 45: Being added or simply integrated in the sense that `TORCH_CHECh` is an RuntimeError, `TORCH_WARN` is a basic warning, `TORCH_CHECK_TYPE` is a TypeError, etc

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,103 @@
...
+    so that will need to be added
+
+* Creating new message classes and severity levels should be easy
+
+* Settings to turn warnings for a specific message class into errors
```
albanD, line 50: Should that be different from the python CLI/warnings module options?

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,103 @@
...
+* Settings to disable specific message classes and severity levels
+
+  - TODO: However, most errors should not be disableable, right? Perhaps only
+    some message classes should allow disabling or downgrading errors. For
+    instance, currently in PyTorch, we can downgrade a nondeterministic error
+    to a warning, but we wouldn't want to downgrade an error from invalid
+    arguments given to an operation.
+
+* Settings to avoid emitting duplicate messages generated by multiple
+  `torch.distribted` ranks (related to issue
```
albanD, line 61: nit: typo

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,103 @@
...
+
+## **Motivation**
+Original issue: [link](https://github.com/pytorch/pytorch/issues/72948)
+
+Currently, it is challenging for PyTorch developers to provide messages that
+act consistently between Python and C++.
+
+It is also challenging for PyTorch users to manage the messages that PyTorch
+emits. For instance, if a PyTorch user happens to be calling PyTorch functions
+that emit lots of warnings, it can be difficult for them to filter out those
```
albanD, line 101: python warning filter should already work for that?

albanD (commented):
Thanks for writing this down!
For the warning/error case, I'm curious what is the discrepancy / missing part in the current system?

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,103 @@
...
+
+* Ability to make a particular warning only warn once
+
+  - NOTE: Currently `TORCH_WARN_ONCE` does this in C++, but there is no Python
+    equivalent
+
+  - TODO: Should there be a setting to turn a warn-always into a warn-once for
+    a given message class and vice versa?
+
+  - TODO: Should warn-once be its own separate severity level?
```
mruberry, line 72: Probably not -- the great majority of warnings should probably be warn-once, the only case where warning every time is reasonable is if the precise reason for the warning is actually differing

RFC-0026-logging-system.md
mruberry: Using existing Python features wherever possible is a great idea (although it doesn't help C++ API users, but we could choose not to support them at this time). We should provide documentation for how to handle common scenarios either way. I'm not sure how Python warning filters work with various logging systems.

mruberry:
Seems like a good proposal in general, @kurtamohler. I think it'd be helpful to expand this to the next level of detail and show how some of these scenarios would work, and how errors and warnings can be originated in both Python and C++

RFC-0026-logging-system.md
albanD: I don't think the two are exclusive:
If the "catching" code is in python, then python-provided filters should work.
If the "catching" code is in c++, there is no existing system and we could build one (I would agree with @mruberry though that this is most likely low priority).

In any case, all warning thrown from python or c++ can be caught in python.
All warnings thrown from c++ can be caught in c++.

kurtamohler: Makes sense, the Python warning filter seems good enough.

However, on a related note, it would also be good to have a way to filter out **info** severity messages, which are just `print` calls at the moment. Related to https://github.com/pytorch/pytorch/issues/68768

RFC-0026-logging-system.md
kurtamohler: The warnings module should be good enough. I'll just change this to say that our warnings must be able to be turned into errors with the warnings module. I guess we'll probably get that behavior for free, but I think it's worth mentioning in the RFC regardless--I imagine we'll probably eventually want a unit test for it

RFC-0026-logging-system.md
kurtamohler: That may be the right idea.

However, I don't think I completely understand what @mruberry meant by "classes of messages" in the description of https://github.com/pytorch/pytorch/issues/72948. I was thinking that class and severity would be independent of each other--every message class can have any severity level. But if, for example, we say that `TypeError` is one of the message classes, then the only possible severity level for it would be Error

kurtamohler: I might have been overthinking this. Maybe it's actually totally fine for a particular message class to only have one severity level that it can be invoked with. In that case, the definition of "message class" could include each of the classic Python errors and warnings, plus any others that we would like to add, like nondeterministic/beta/prototype messages.

Still, I'm not completely sure what @mruberry had in mind

kurtamohler:
Would it be at all helpful if I include a somewhat full description (or links to documentation when possible) for the current message APIs in PyTorch? I am going to write notes on it regardless for my own reference, and I can include it here if it's a good idea

RFC-0026-logging-system.md
kurtamohler: Sorry for my confusion, I'm pretty sure I now understand what a "message class" is.

For each severity level, several different message classes can be used. A given message class is only used for one particular severity level. (So my thought that severity and class are independent was wrong.)

For example, with the error severity level in C++, we have the error class `c10::Error` and all of its subclasses defined [here](https://github.com/pytorch/pytorch/blob/72e4aab74b927c1ba5c3963cb17b4c0dce6e56bf/c10/util/Exception.h#L195-L236). When a C++ error propagates to Python, it turns into a corresponding Python error class. For instance `c10::Error` becomes `RuntimeError`, `c10::IndexError` becomes `IndexError`, etc.

At the moment, it looks like only one warning class exists in C++ (although I think it's not actually implemented as a C++ class), and it always propagates to the same Python class, `Warning`. I think one of the goals of this message logging system is to create different warning classes in C++ that propagate to a corresponding Python warning class--much like how error classes currently work--and to also do the same for info severity messages.

So I think the answer to @albanD's question "Should we include all of the classic python error/warnings here?" is probably yes. I'll update the requirements to reflect all of this

kurtamohler:
I've added a lot more detail, including a description of how message logging is currently done in PyTorch, and fixed some of the things brought up in discussion so far.

I haven't described all of the APIs for creating messages in the new system yet--I will do that next

mruberry:
> I've added a lot more detail, including a description of how message logging is currently done in PyTorch, and fixed some of the things brought up in discussion so far.
> 
> I haven't described all of the APIs for creating messages in the new system yet--I will do that next

Hey @kurtamohler! FYI I'll be on PTO for the next several weeks, and I think @albanD is on PTO currently. So this may take some time to respond to.

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+## Implementation details
+
+### C++ to Python Error Translation
+
+`c10::Error` and its subclasses are translated into their corresponding Python
+errors [in `CATCH_CORE_ERRORS`](https://github.com/pytorch/pytorch/blob/72e4aab74b927c1ba5c3963cb17b4c0dce6e56bf/torch/csrc/Exceptions.h#L54-L100).
+
+However, not all of the `c10::Error` subclasses in the table above appear here.
```
albanD, line 449: Most likely oversight when these errors were added or when that macro was made :D

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+I'm not sure yet what's up with that.
+
+`CATCH_CORE_ERRORS` is included within the `END_HANDLE_TH_ERRORS` macro that
```
albanD, line 452: nit: maybe mention that we have a similar macro that should be used for pybind-based bindings called `END_HANDLE_TH_ERRORS_PYBIND`?

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+* `torch::ValueError`
+* `torch::NotImplementedError`
+* `torch::AttributeError`
+* `torch::LinAlgError`
+
+
+### C++ to Python Warning Translation
+
+The conversion of warnings from C++ to Python is described [here](https://github.com/pytorch/pytorch/blob/72e4aab74b927c1ba5c3963cb17b4c0dce6e56bf/torch/csrc/Exceptions.h#L25-L48)
+
```
albanD, line 475: Is there anything notable about the current `Info` system? it is the one I'm least familiar with and I am a bit curious :D

albanD (commented):
That looks great! Thanks for taking the time to write all of this down

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+### Configurability and filtering
+
+* Ability to turn warnings into errors. This is already possible with the
+  Python `warnings` module filter, but the PyTorch docs should mention it and
+  we should probably have unit tests for it.
+  See [documentation](https://docs.python.org/3/library/warnings.html#the-warnings-filter)
+
+* Settings to disable specific message classes and severity levels
+
+  - TODO: Error classes should never be disableable, right?
```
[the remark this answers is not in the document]
rgommers: AFAIK there is no good reason anymore for the `python -OO` mode; it strips not only asserts but also docstrings, in an attempt to save non-essential space (which probably was useful 20 years ago).

There's one web framework that still launches with `-OO`, which is a recurring source of issues in numpy/scipy; any manipulation of `__doc__` should be guarded because `__doc__` can be `None`. It's a hassle and fairly pointless.

> So I would vote for no.

me too

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+
+### Message APIs
+
+In order to emit messages, developers can use the APIs defined in this section.
+
+These APIs all have a variable length argument list, `...` in C++ and `*args`
+in Python. When a message is emitted, these arguments are concatenated into
+a string, and the string becomes the body of the message. In C++, the arguments
+must all have the `std::ostream& operator<<` function defined so that they can
+be concatenated, and in Python, they must all have a `__str__` function.
```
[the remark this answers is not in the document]
kurtamohler: Right, in C++, `very_expensive_function()` won't be called if `cond == true`. Same for `WARN_ONCE(very_expensive_function())`, if it already has warned once before. I'll add that detail to the doc.

I might as well ask, even though I'm almost sure the answer is no--is there any possible way to avoid calling `very_expensive_function()` in Python as well for `torch.check(true, very_expensive_function())`?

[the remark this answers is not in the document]
albanD: > I might as well ask, even though I'm almost sure the answer is no--is there any possible way to avoid calling very_expensive_function() in Python as well for torch.check(true, very_expensive_function())?

I don't think so indeed. Maybe we want to consider that this function takes a callable as the second argument to preserve this property?
Or maybe this is ok and we will rely on Dynamo to get this behavior cc @ezyang ?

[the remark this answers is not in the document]
ezyang: Just pass it in a lambda. This is how torch._prims_common.check works

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+
+  - Filtering out **Info** severity messages would also be nice to have, since
+    excessive printouts can degrade the user experience. Related to
+    issue [#68768](https://github.com/pytorch/pytorch/issues/68768)
+
+* Settings to enable/disable emitting duplicate messages generated by multiple
+  `torch.distributed` ranks. Related to issue
+  [#68768](https://github.com/pytorch/pytorch/issues/68768)
+
+* Ability to make a particular warning only warn once. Warn-once should be the
```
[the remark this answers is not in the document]
kurtamohler: That seems like a good idea

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+
+* **`c10::DeprecationWarning`** - Python `DeprecationWarning`
+  - Emitted when a deprecated function is called.
+  - TODO: `DeprecationWarning`s are ignored by default in Python, so we may
+    actually want to use a different Python class for this.
+
+
+#### Info message classes:
+  
+* **`c10::Info`** - Python `torch.Info`
```
albanD, line 191: Thinking more about what would be the use case for this, I don't think that this should be used within core in general.
The only use case that I could see is a particular submodule that is expected to do some logging (distributed runner, benchmark tools, etc).
What do you think about forbidding to use the base `Info` class and you are only allowed to use subclasses (DistributedInfo, BenchmarkInfo, etc). That will reduce the chances of having unrelated logs interleaved that the user cannot silence easily.

kurtamohler: I like that idea

edward-io:
we have torch.monitor (https://github.com/pytorch/rfcs/pull/30) that hasn't seen too much usage, perhaps we can use some of it for the new logging system.

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+callsite. It is not possible to implement it this way in Python, so need to
+think of some other way to do it. Of course the Python `warnings` module's
+[`"default"` filter](https://docs.python.org/3/library/warnings.html#the-warnings-filter)
+prevents duplicate warnings from being emitted, but it acts a little
+differently--if two warning messages emitted from the same location differ even
+slightly (for instance, if the value of some variable is included in the
+message and that value differs between two different `warnings.warn` calls),
+then both warnings are emitted. `TORCH_WARN_ONCE` does not check whether
+messages differ. But we could probably implement `torch.warn_once` in a similar
+way to how the `warnings` module filter is implemented.
```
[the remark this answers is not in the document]
kurtamohler: Yep, that's basically how they do it: https://github.com/python/cpython/blob/8a7bf2d28cdb3579cd33a9c9deb9b5d30eacb086/Lib/warnings.py#L375

The key for the map is a tuple containing the message body, warning class type, and line number: https://github.com/python/cpython/blob/8a7bf2d28cdb3579cd33a9c9deb9b5d30eacb086/Lib/warnings.py#L346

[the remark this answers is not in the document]
kurtamohler: I guess we could potentially create something like a "once-location" filter that acts similarly to the "once" filter in the `warnings` module, except that it only pays attention to the warning class, line number, and filename, ignoring the body of the message

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+  * The warning is only emitted if `my_rank == warn_rank`
+
+TODO: Add APIs for the rest of the message classes, like
+`torch.log_info_rank()`, etc.
+
+TODO: There should also be a global setting to enable emitting the duplicates.
+`torch.warn_rank` could check the setting, and if it's turned on, then it would
+emit the warning for all ranks.
+
+TODO: Should we have a `TOCH_WARN_RANK` (and others) in C++ as well? Is there
```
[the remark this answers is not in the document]
kumpera: This is better done by having a format string that is used to define how log lines are produced.

kumpera:
How is one supposed to selectively change the log verbosity of a specific subsystem?
And how would a subsystem author produces logs that are properly scoped?

While having functions like `torch.log_info` is nice, it would be great if we could construct logger instances for particular modules. For example:

```python
# distributed.py

logger = torch.Logger("torch.distributed")

def broadcast():
 logger.log_info("broadcasting stuff")
```

Finally, is there a reason on why we can't integrate with python's logging package so there's very little users need to do to leverage this?

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+  
+
+#### Warning message classes:
+
+* **`c10::UserWarning`** - Python `UserWarning`
+  - Default warning class. Other warning classes inherit from it.
+
+* **`c10::BetaWarning`** - Python `torch.BetaWarning`
+  - Emitted when a beta feature is called. See
+    [PyTorch feature classifications](https://pytorch.org/blog/pytorch-feature-classification-changes/).
```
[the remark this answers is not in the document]
kurtamohler: Yeah, I can see your point, maybe beta/prototype warning classes aren't a good idea. There are some warnings currently being emitted in these cases, for instance: https://github.com/pytorch/pytorch/blob/56dea92d973ff16b41cb812b69c5a7ab8c9d1109/torch/csrc/cuda/Module.cpp#L566-L568

Should we just use the base `c10::Warning` class for beta/prototype features that currently emit a warning, or provide something that's just a little more specific, like maybe `UnstableFeatureWarning`?

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,481 @@
...
+warnings, the following signature could be used:
+
+**`torch.warn_rank(my_rank, *args, warn_rank=0)`**
+  * Args:
+    - `my_rank` - Rank of the subprocess calling this function
+    - `args` - Warning message
+    - `warn_rank` - Rank that should emit the message
+  * The warning is only emitted if `my_rank == warn_rank`
+
+TODO: Add APIs for the rest of the message classes, like
```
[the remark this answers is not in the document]
kurtamohler: Yeah, you're probably right. Checking `torch.distributed.get_rank()` within `torch.warn()` should probably be good enough

RFC-0026-logging-system.md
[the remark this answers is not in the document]
kurtamohler: @albanD, yeah I think we could make this work by default. I'm not sure if an API for checking the distributed rank currently exists in C++ (I looked briefly and didn't find it), but if not, I don't think there's any reason why we couldn't add it.

As for having all ranks send warnings/errors to rank 0, that would be interesting. I'm not sure how to implement that. Also, would sending all warnings to rank 0 

@kumpera, I'm not sure what you mean. Are you saying that you can use a format string to filter out duplicate messages from different ranks? Can you give an example?

kurtamohler:
@kumpera, good questions.

> How is one supposed to selectively change the log verbosity of a specific subsystem?

At the moment, the general idea is that users can use a filter to silence messages. The specific API for that hasn't been written down yet. Although, in the case of warnings, the `warnings` module already does this. The missing detail is how the user would silence info messages, and I imagine that we would want to have something that has a very similar interface to the `warnings` module filter.

> And how would a subsystem author produces logs that are properly scoped?

Authors would use the appropriate message class. If an applicable message class doesn't exist yet, they should add it.

> While having functions like `torch.log_info` is nice, it would be great if we could construct logger instances for particular modules.

That might be a good idea. What do you think @albanD, @mruberry ?

> Finally, is there a reason on why we can't integrate with python's logging package so there's very little users need to do to leverage this?

I don't know much about it, I'll read about it and get back to you

RFC-0026-logging-system.md
```diff
@@ -0,0 +1,486 @@
...
+  - C++ error: `c10::TypeError`
+  - Python error: `TypeError`
+
+**`TORCH_CHECK_NOT_IMPLEMENTED(cond, ...)`** - `torch.check_not_implemented(cond, *args)`
+  - C++ error: `c10::NotImplementedError`
+  - Python error: `NotImplementedError`
+
+**`TORCH_CHECK_WITH(error_t, cond, ...)`** - `torch.check_with(error_type, cond, *args)`
+  - C++ error: Specified by `error_t` argument
+  - Python error: Specified by `error_type` argument
```
zou3519, line 244: We want to change how asserts are written to support vmap (and/or Tensor Subclasses) over functions that call asserts (e.g. https://github.com/pytorch/pytorch/pull/81767, cc @kshitij12345). In order to support those, the assert should go through the PyTorch dispatcher. Since we are re-thinking how asserts in PyTorch work, this is a good opportunity to figure out how we can have asserts as PyTorch operators.

## Some background
folks write functions in C++ like the following:
```
Tensor safe_log(const Tensor& x) {
  TORCH_CHECK((x > 0).all(), "requires x > 0");
  return x.log()
}
```
If someone is doing `vmap(safe_log)(torch.tensor([-1, 1, 2))`, then `(x > 0).all()` is a Tensor that has multiple values. A Tensor that has multiple values cannot be coerced to a single boolean value, and the above errors out.

## A sketch of a solution

For synchronous asserts, we would want two extension points. One to actually check the condition, and the other to raise an error message. So `TORCH_CHECK(cond, ...)` would be implemented like:
```
if (at::is_true(cond)) {
  msg = ...
  at::fail(msg)
}
```

kumpera:
> 
> > And how would a subsystem author produces logs that are properly scoped?
> 
> Authors would use the appropriate message class. If an applicable message class doesn't exist yet, they should add it.
> 

Sorry for not being specific here, what I meant by scoping is which subsystem is producing a given log message.

For example, if one is troubleshooting an issue on backwards of a distributed model, they would enable logging for the "autograd" and "distributed" subsystems.

I guess this boils down to whether log messages would carry a subsystem tag that can be used as part of filtering.

RFC-0026-logging-system.md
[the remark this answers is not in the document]
kumpera: @kurtamohler the format of each line should be controlled by a format string.  From python's logging: https://docs.python.org/3/library/logging.html#logging.Formatter

An user dealing with logging of a distributed application could use something like: "[%(rank)] (%(time)) - %(message)" to produce a line like "[0] (2022-01-01 00:01:00) - distributed failed for some reason"

@albanD rank is available by calling `torch.distributed.get_rank()`. It's a bit more involved from C++.

On the subject of log shipping, providing a default one is a great idea as it would improve the OOTB experience.

RFC-0026-logging-system.md
kurtamohler: @zou3519, I started looking into adding vmap support to `TORCH_CHECK`, and I realized that it doesn't actually support tensor inputs at all, not even non-batched tensors. For example, if I do this:

```cpp
TORCH_CHECK((at::zeros(10).eq(0)).all());
```

I get this build error:

```
/work2/kurtamohler/development/pytorch-1/c10/util/Exception.h:506:29: error: no match for 'operator!' (operand type is 'at::Tensor')
  506 |   if (C10_UNLIKELY_OR_CONST(!(cond))) {            \
```

Without `at::Tensor::operator!`, we currently have to do something like this instead:

```cpp
TORCH_CHECK((at::zeros(10).eq(0)).all().item<bool>());
```

So I like the idea of adding support for tensor inputs to `TORCH_CHECK`.

One way, could be to use an `at::is_true` function, like you mentioned. However, I'm not sure exactly how to do that. Since `TORCH_CHECK` is defined in `c10`, it doesn't have access to the `at` namespace. But of course, since it's a macro, it's just doing substitution, so maybe we could have an `#ifdef` that checks whether `at::is_true` is available before using it, like this:

```cpp
#ifdef AT_IS_TRUE_DEFINED
#define CHECK_COND(cond) at::is_true(cond)
#else
#define CHECK_COND(cond) (cond)
#endif
```

The `#define AT_IS_TRUE_DEFINED` would be next to wherever the declaration of `at::is_true` is, and then we can use `CHECK_COND` in the definition of `TORCH_CHECK`.

But when I tried this, I couldn't figure out which file to define `at::is_true` such that its available for any `TORCH_CHECK` call that might use tensors. It also just seems like kind of a sketchy solution.

Another option is to just define `at::Tensor::operator!`. But in order to make `TORCH_CHECK` work correctly when `cond` is a batched tensor, the actual meaning of `at::Tensor::operator!` would have to be like `!physical_tensor.any()`. That seems like an inconsistency to me, but maybe it's actually fine, I'm not sure.

Maybe there's another better way to do this.

ezyang: Is there any reason you have to use TORCH_CHECK for this

kurtamohler: As I understood it, the suggestion was to make `TORCH_CHECK((x > 0).all(), "requires x > 0");`, for example, work for both batched and non-batched tensors

ezyang: I guess, for something like this where we're syncing, I think it is fine if we force the call sites to change what they look like. There's not many of this kind of check.

zou3519: I'm fine with having something like a TORCH_CHECK_TENSOR that accepts a Tensor to sidestep the issues above. We control what gets written in our C++ internals so we can always refactor if we're not satisfied.

In Python, is it correct that we want `torch.check` to be public API? If so it needs more scrutiny. My initial thought is that torch.check should do both TORCH_CHECK and TORCH_CHECK_TENSOR (there aren't technical reasons in Python for us to split those up)

kurtamohler: I don't think the Python API needs to be public. I can make it private to start with, and we can make it public if we ever find a reason to

PR #53, opened 2023-04-18
RFC-0030 Consolidate TorchElastic and TorchX

kiukchung:


kiukchung:
cc @soumith, @msaroufim, @d4l3k, @kurman, @priyaramani

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,262 @@
...
+[TorchX](https://pytorch.org/torchx)
+on the other hand is a **job launcher** that can submit PyTorch scripts as a job onto various remote schedulers
+such as SLURM, Kubernetes, AWS/GCP/Azure Batch, etc. Similar to TorchElastic prior to `torch-1.9`,
+TorchX is currently in the PyTorch GitHub organization but under the [pytorch/torchx](https://github.com/pytorch/torchx)
+repository.
+
+
+## **Motivation**
+
+1. **Uniform user experience (UX)**
```
msaroufim, line 76: The motivations would lead me to believe that either of the below are valid options
1. Moving elastic out of core into torchx
2. Moving torchx into core with elastic 

At a high level I agree that the needs served by torchrun and torchx are quite similar so should be coupled more closely. One worry I have is that typically in core they're strict about adding new dependencies and often prefer out of core registration mechanisms

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,262 @@
...
+such as SLURM, Kubernetes, AWS/GCP/Azure Batch, etc. Similar to TorchElastic prior to `torch-1.9`,
+TorchX is currently in the PyTorch GitHub organization but under the [pytorch/torchx](https://github.com/pytorch/torchx)
+repository.
+
+
+## **Motivation**
+
+1. **Uniform user experience (UX)** 
+     1. Both `torchrun` and `torchx` **launch** PyTorch scripts 
+     2. For instance, `torchx run -s local dist.ddp -j 1x2 train.py` is basically equivalent to `torchrun --nnodes 1 --nproc_per_node 2 train.py`
```
msaroufim, line 78: this makes me feel that maybe torchrun is actually the better name, it's more descriptive

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,262 @@
...
+
+See section on BC above.
+
+## **Alternatives**
+
+We considered pulling `torch.distributed.elastic` (torchelastic) out to the pytorch/torchx repository and
+consolidate TorchElastic and TorchX *outside* the pytorch/pytorch repo. However, due to the prevalent usage
+of `torchrun`, pulling torchelastic out of PyTorch would mean:
+
+1. Makes `torch-2.x` backwards incompatible since users would have to separately install `torchx` to get `torchrun`.
```
msaroufim, line 232: I don't feel like these drawbacks are particularly bad

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,262 @@
...
+   Once migrated over to mypy the regular PyTorch [Lint GitHub action](https://github.com/pytorch/pytorch/blob/main/.github/workflows/lint.yml) should suffice.
+2. **Conditional Integration Tests**: TorchX runs two types of end-to-end integ tests to ensure that it can submit jobs correctly to the supported schedulers:
+     1. [Components Integration Test](https://github.com/pytorch/torchx/blob/main/.github/workflows/components-integration-tests.yaml):
+        Runs locally on the CI runner by using mock/local schedulers (e.g. minikube for kubernetes)  
+     2. [SCHEDULER] Integration Tests (e.g. [aws-batch-integration-tests](https://github.com/pytorch/torchx/blob/main/.github/workflows/aws-batch-integration-tests.yaml)):
+       Runs against scheduler instances running on an actual cloud (e.g. aws, gcp) account.
+   
+   While (i) is lighter weight compared to (ii) both are rather expensive to run at the pace of commits to PyTorch.
+   Therefore, we propose that these integ tests are run:
+     1. Only for commits that touch files in `torch/x/**`
```
msaroufim, line 213: this is what inductor does for example and makes sense to me

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,262 @@
...
+     1. In practice running large-scale jobs reliably requires coordination between the scheduler/infrastructure (controlled by torchx)
+         and the process manager (torchelastic)
+     2. Example: dealing with host failures in a job:
+        1. torchelastic alone cannot deal with host failures. The job needs to be configured (done via torchx) for
+           the scheduler to deal with host failures (e.g. replace the failed host).
+        2. Conversely, the scheduler replacing failed hosts for a job is not enough, the job itself needs to be authored
+           in such a way that surviving hosts can wait-for a failed host to be replaced then admit the newly replace host
+           into the worker group (provided by torchelastic).
+3. **Out-of-the-box support for job launchers in PyTorch** 
+    1. Without additional installs, PyTorch users can run PyTorch scripts both locally and remotely onto widely used schedulers
```
msaroufim, line 89: Actually if this is all living in core it might make sense to have all schedulers out core with maybe some exception for the very well maintained or popular ones like K8, that way you can standardize a job launching interface without having to contribute more experimental schedulers in core where they may likely rot

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,262 @@
...
+     2. Write each sub-proc's stdout and stderr to log files (configurable)
+     3. Prefix each line of stdout and stderr with local rank information
+     4. Propagate exception traces from sub-proc back to the main proc for better error handling + propagation
+
+### Backwards Compatibility (BC)
+
+#### non-BC Breaking Changes
+
+1. `torch/distributed/launch.py` and `torch/distributed/run.py` would have to remain for a few releases,
+   but will both point to `torch/x/run.py`.
```
mrshenli, line 126: The consolidation makes a lot of sense to me. Do you and your plan plan to support this change all the way till `torch/distributed/launch.py` and `torch/distributed/run.py` are eventually removed from the repo?

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,262 @@
...
+#### BC Breaking Changes
+
+The use-cases below will be BC breaking under this proposal. For each item, we discuss ways to make them non-BC breaking.
+However we propose that we clearly document the migration process and add references to these docs in the release notes
+in favor of introducing code complexity and/or tech-debt to keep things BC.
+
+
+1. Programmatic usages of torchelastic (e.g. `import torch.distributed.elastic`) are **NOT** BC and the user has to codemod
+   references to `torch.distributed.elastic` to `torch.x.elastic`.
+    1. **Impact**: Besides Meta-internal use-cases (which can be easily codemoded) the only external programmatic usage
```
mrshenli, line 144: no problem, we can do the codemod for internal use-cases.

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,262 @@
...
+        of torchelastic is in DeepSpeed (see [GitHub](https://github.com/microsoft/DeepSpeed/search?q=torch.distributed.elastic))
+        for which we can work with the project owner to resolve.
```
mrshenli, line 146: Are we confident this is the only external usage? I do recall seeing many users asking questions about TorchElastic/TorchX on forum.

RFC-0030-Consolidate-Elastic-and-TorchX.md
kiukchung: yea, I don't want to bring additional deps + unnecssary bloat to core either. However there are number of benefits for PyTorch to define the specifications of how PyTorch authored scripts/applications are installed and run on the target infrastructure:

1. PyTorch can intentionally ask for specific features and requirements from the infra to define the runtime environment to its liking (e.g. a flash storage side-car for checkpointing, node-level retries from schedulers for fault-tolerance, specific network topologies for certain workloads) 
2. Helps & encourages the infra/platform providers to support PyTorch in the way PyTorch is natively designed to be run.
3. 1) and 2) consequentially makes the PyTorch UX more uniform, portable, and easy across different infra providers.

RFC-0030-Consolidate-Elastic-and-TorchX.md
kiukchung: yea I think it makes sense to include the schedulers that are not CSP specific (which would be SLURM and Kubernetes - SLURM doesn't even have a build-time dep since it doesn't have a python client), and keep the CSP ones (e.g. AWS Batch) in torchx so that eventually we can make the relevant service teams own the integration.

RFC-0030-Consolidate-Elastic-and-TorchX.md
kiukchung: Yep that's the plan. Ideally we'd like to keep `torchrun` still in pytorch (which is what this RFC is advocating for) but `torch/distributed/launch.py | run.py` would be moved out to `torch.x.run`

RFC-0030-Consolidate-Elastic-and-TorchX.md
kiukchung: AFAICT programmatic usage + more well known 3rd party lib yes. Most people use `torchrun` or `python -m torch.distributed.launch`. Even the kubeflow pytorch op uses `python -m torch.distribued.run` (https://github.com/kubeflow/training-operator/blob/master/examples/pytorch/elastic/imagenet/imagenet.yaml)

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,321 @@
...
+    1. **To make BC**: We could re-import `torch.x.elastic.**` from `torch/elastic/**/__init__.py`. For example,  
+        ```python
+        # In torch/distributed/elastic/__init__.py
+        import torch.x.elastic.* # noqa F401
+        ```
+2. (OSS-only) `torchx` CLI users would have to switch over to `torchrun`.
+    1. **Impact**: Every OSS user that currently uses `torchx` CLI would have to one-time opt-in and switch over to torchrun
+    2. **To make BC**: 
+         1. **(Option 1)** Add console script `torchx` to PyTorch's `setup.py`, which would act as a symlink discussed in the non-BC section for Meta-internal use-case
```
priyaramani, line 207: I like the symlink option to keep disruptions for users minimal but we need to migrate at some point

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,321 @@
...
+              1. Effectively splits the TorchX repo in two hence makes maintenance and CI/CD (testing) more complex
+              2. No built-in support for launching remote jobs
+     2. **[ PREFERRED ] (Option 2) Option 1 + Kubernetes and SLURM support**
+          1. PROS: Keeps the minimalistic approach of Option 1 while providing out-of-the-box for CSP agnostic remote scheduler support
+          2. CONS: TorchX still split into two repos
+         
+        > NOTE: The rest of the doc assumes Option 2 as this is the recommended option
+     3. (Option 3) Upstream everything in TorchX:
+          1. PROS: Makes maintenance and CI/CD simple since changes to the specs can be tested for downstream BC easily
+          2. CONS: bloat-torch
```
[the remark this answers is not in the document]
kiukchung: Right so regardless of which option, the "extra-deps" of torchx would **never** be added as deps (even extra deps) of `torch`. **We won't mess with the `extras_require` of `torch`** (aka no `torch[batch]`) and instead would fail-fast with a warning prompting the user to install additional packages.

RFC-0030-Consolidate-Elastic-and-TorchX.md
```diff
@@ -0,0 +1,321 @@
...
+3. Merge the functionalities of `torchx` CLI to
```
osalpekar, line 160: Consolidating around `torchrun` sounds like a good idea. My understanding from the above is that:
* `torchrun` would now take a scheduler arg, and users can launch on Slurm/Kubernetes out of the box with just the PyTorch package. 
* Other schedulers will be implemented in the `torchx` repo, and users would need to install torchx + do some registration to use those schedulers.
* Runtime libraries like artifact tracking will remain in `torchx`, as will non-runtime stuff like pre-defined components, the pipelines API, etc.

With this approach what will be the framing for the torchx repo? It may come across as a set of disparate pieces that might be useful during training.

kiukchung: > `torchrun` would now take a scheduler arg, and users can launch on Slurm/Kubernetes out of the box with just the PyTorch package.

Yes exactly. There's a bit of tricky BC to deal with in terms of CLI arguments (`torchrun` and `torchx` CLIs currently have completely different args - `torchx` focuses more on actions that can be done at the job level whereas `torchrun` focuses on a "single action": launching multiple procs on the local node).

> Other schedulers will be implemented in the torchx repo, and users would need to install torchx + do some registration to use those schedulers.

The idea is to upstream the `torchx.specs.*` (iface and APIs) to `torch.x.*` and optionally (but strongly encouraged) upstream 1 non-trivial scheduler integration. IMO SLURM is a good candidate since it has no "client" dependencies (e.g. `torchx.schedulers.kubernetes` uses the [`kubernetes`](https://pypi.org/project/kubernetes/) python client). The existing schedulers (k8s, aws-batch, gcp-batch, azure-batch, ray) should stay in the torchx repo as an optional plugin.

As for registration of the schedulers, this is already possible in torchx by design (see [Registering Custom Schedulers](https://pytorch.org/torchx/latest/advanced.html#registering-custom-schedulers)). In fact this is how we currently make only the internal schedulers visible to Meta-internal uses.


> Runtime libraries like artifact tracking will remain in torchx, as will non-runtime stuff like pre-defined components, the pipelines API, etc.

Yes for now. `torchx.runtime.*` (including artifact tracking) APIs are currently not stable therefore, I think should be excluded from the upstream. The pipeline integrations should be deprecated (since no one seems to be using it). As for the pre-def components, I think its useful to have `python` and `ddp` (probably should be renamed to `spmd`) in the main repo since those are the two most used launch topologies. They would serve two purposes: 

#. As a vanilla/basic quick start components to launch test and PTD jobs
#. As reference implementations to further customize those components to fit your exact needs/infra.

> With this approach what will be the framing for the torchx repo? It may come across as a set of disparate pieces that might be useful during training.

IMHO the TorchX repo should serve as a place to hold commonly used plugins (e.g. widely used scheduler impls, most commonly used resources (host types), etc).

PR #54, opened 2023-04-28
NumPy support in TorchDynamo

lezcano:
This RFC proposes a design and implementation for a translation layer from NumPy to PyTorch. This is currently implemented at [numpy_pytorch_interop](https://github.com/Quansight-Labs/numpy_pytorch_interop/).

We leave an open question in the section "Unresolved Questions" regarding whether this API should also be public or not.

### DIFF
--- a/RFC-0032-numpy-support-in-dynamo.md
+++ b/RFC-0032-numpy-support-in-dynamo.md
@@ -198,6 +198,11 @@ layer does not implement this new API, as the default bit generator in NumPy is
 From this, it follows that this API will not give any reproducibility
 guarantees when it comes to randomness.
 
+**Accuracy**. For deterministic operations, we would expect to give accuracy
+guarantees similar to those in PyTorch 2.0. In particular, we would expect
+these decompositions to be as precise as those from NumPy when compared to an
+`fp64` baseline minus perhaps a small relative error.
+
 
 ## The `torch_np` module
 
--- a/RFC-0032-numpy-support-in-dynamo.md
+++ b/RFC-0032-numpy-support-in-dynamo.md
@@ -1,0 +1,380 @@
+# A PyTorch - NumPy compatibility layer
+
+**Authors:**
+* @ev-br
+* @lezcano
+* @rgommers
+
+## Summary
+This RFC describes a proposal for a translation layer from NumPy into PyTorch.
+In simple terms, this accounts for implementing most of NumPy's API (`ndarray`,
+the `numpy`, `numpy.linalg`, `numpy.fft`  modules, etc) using `torch.Tensor`
+and PyTorch ops as backend.
+
+The main goal is: **make TorchDynamo understand NumPy calls**.
+This should enable an end user to combine code that uses the PyTorch API with
+code that uses the NumPy API, in a way that allows TorchDynamo to understand
+those function calls and build up an execution graph. To enable this, it is key
+that there is a translation layer from NumPy to PyTorch function calls, which
+TorchDynamo can use in order to build up its execution graph from PyTorch
+functions/primitives only. For niche functions in NumPy that don’t have a
+PyTorch equivalent, it’s okay to graph break and still call NumPy to execute
+the function call.
+
+The work is currently being done at [numpy_pytorch_interop](https://github.com/Quansight-Labs/numpy_pytorch_interop/).
+
+
+## Motivation
+
+### Introductory examples
+
+Consider the following snippet:
+```python
+import numpy as np
+
+x = np.random.randn(3, 4)
+y = np.random.randn(4, 3)
+z = np.dot(x, y)
+w = z.sum()
+```
+
+When we trace this program with the compat layer, the semantics of the
+program would stay the same, but the implementation would be equivalent to
+
+```python
+import torch
+x = torch.randn(3, 4, dtype=torch.float64)
+y = torch.randn(4, 3, dtype=torch.float64)
+z = torch.matmul(x, y)
+w = z.sum()
+```
+
+Here, we can already spot a couple differences between NumPy and PyTorch.
+The most obvious one is that the default dtype in NumPy is `float64` rather than
+`float32`. The less obvious is very sneakily hiding in the last line.
+
+```python
+>>> type(w)
+<class 'numpy.float64'>
+```
+
+Reductions and similar operations in NumPy return the infamous NumPy scalars.
+We'll discuss these and other NumPy quirks and how we dealt with them in the
+[design decision section](#design-decisions).
+
+
+Let's now have a look at a toy example of how this layer would be used.
+```python
+import torch
+import numpy as np
+t1 = torch.tensor([1, 3, 5])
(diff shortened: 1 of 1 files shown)


RFC-0032-numpy-support-in-dynamo.md
```diff
@@ -0,0 +1,380 @@
...
+### Limitations
+
+A number of known limitations are tracked in the second part of the
+[OP of this issue](https://github.com/Quansight-Labs/numpy_pytorch_interop/issues/73).
+When landing this RFC, we will create a comprehensive document with the differences
+between NumPy and `torch_np`.
+
+### Beyond plain NumPy
+
+**GPU**. The current implementation so far only been implemented and tested on
```
albanD, line 340: It would be great to have some code sample on what you have in mind here? and in particular in eager vs compiled mode? and when using explicit bridge vs `__array__` based bridge.

RFC-0032-numpy-support-in-dynamo.md
```diff
@@ -0,0 +1,380 @@
...
+## Unresolved questions
+
+A question was left open in the initial discussion. Should the module
+`torch_np` be publicly exposed as `torch.numpy` or not?
+
+A few arguments in favor of making it public:
+* People could use it in their NumPy programs just by changing the import to
+  `import torch.numpy as np`. This could be a selling point similar to JAX's
+  `jax.numpy`, which could incentivize adoption.
+* People would not need to use the whole PyTorch 2.0 stack to start using
```
albanD, line 366: Not sure what this sentence means?

RFC-0032-numpy-support-in-dynamo.md
```diff
@@ -0,0 +1,380 @@
...
+def fn(x, y):
+    return np.multiply(x, y).sum()
+
+result = fn(t1, t2)
+t_results = torch.empty(5, dtype=torch.float64)
+t_results[0] = result  # store the result in a torch.Tensor
+```
+
+Note that this code mixing NumPy and PyTorch already works in eager mode with
+CPU tensors, as `torch.Tensor` implements the `__array__` method. Now, the
```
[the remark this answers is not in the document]
lezcano: I don't know how much is this used in the wild, but it's certainly used in the PyTorch test suit, for example. My guess would be that it's quite used, even if by chance, whenever you mix PyTorch and NumPy, given that if you're dealing with CPU tensors and you forgot to write the `.numpy()` call things just work.

RFC-0032-numpy-support-in-dynamo.md
```diff
@@ -0,0 +1,380 @@
...
+```python
+>>> np.int32(2) * [1, 2, 3]       # scalar decays to a python int
+[1, 2, 3, 1, 2, 3]
+
+>>> np.asarray(2) * [1, 2, 3]     # zero-dim array is an array-like
+array([2, 4, 6])
+```
+
+We don't expect these to pose a big issue in practice. Note that in the
+proposed implementation `np.int32(2)` would return the same as `np.asarray(2)`.
```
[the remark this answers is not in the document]
lezcano: As discussed below, I think this is something we have to live with. `np.sum(x)` will return a `np.float64` and there's very little you can do about it. Our decision was to be consistent all throughout and simplify things by unifying 0D and NumPy scalars. Note that there aren't very many places where this would be an issue. The most problematic issue is the weird broadcasting rules that 0D and NumPy scalars have, but that's being deprecated in NumPy 2.0 under [NEP 50](https://numpy.org/neps/nep-0050-scalar-promotion.html). c.f.
>This NEP proposes to refactor the behaviour around two guiding principles:
> 1. [...]
> 2. NumPy scalars and 0-D arrays should behave consistently with their N-D counterparts.

RFC-0032-numpy-support-in-dynamo.md
```diff
@@ -0,0 +1,380 @@
...
+be simplified, saving us the pain of having to implement all the pre-existing
+corner-cases.
+
+**Randomness**. PyTorch and NumPy use different random number generation methods.
+In particular, NumPy recently moved to a [new API](https://numpy.org/doc/stable/reference/random/index.html)
+with a `Generator` object which has sampling methods on it. The current compat
+layer does not implement this new API, as the default bit generator in NumPy is
+`PCG64`, while on PyTorch we use `MT19937` on CPU and `Philox` on non-CPU devices.
+From this, it follows that this API will not give any reproducibility
+guarantees when it comes to randomness.
```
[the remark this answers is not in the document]
lezcano: Yeah, that'd be the desired outcome, yes. Note that we have not been able to properly test this, as NumPy does not have any sort of system to generate valid inputs for a given operation.

I'll add a line making the point that this is what we really want.

RFC-0032-numpy-support-in-dynamo.md
```diff
@@ -0,0 +1,380 @@
...
+by the end user.
+
+The second step is then done via type annotations and a decorator. Each type
+annotation has an associated function from NumPy-land into PyTorch-land. This
+function converts the set of inputs accepted by NumPy for that argument into a
+PyTorch-land object (think a `torch.Tensor` or a PyTorch dtype). For example,
+for `np.diag` we would write
+
+```python
+def diag(v: ArrayLike, k=0):
```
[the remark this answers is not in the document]
lezcano: Added for readability. For reference, we wrap all the functions with the decorator automatically in a for-loop, that's why I didn't initially added it there.

RFC-0032-numpy-support-in-dynamo.md
lezcano: Updated.

RFC-0032-numpy-support-in-dynamo.md
lezcano: The idea here is that using `torch.compile` already adds a fair amount of technical overhead: when something doesn't work, you get huge errors, sometimes informative, sometimes not. If this API were public, you could first test it just by changing the relevant imports, and if anything breaks, you'll get a readable error. Then, once you know your implementation is sound in eager mode, you can simply add the `torch.compile` decorator and have it sped up

RFC-0032-numpy-support-in-dynamo.md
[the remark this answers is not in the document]
albanD: Yes that sounds good now that I have more context on the description below!

RFC-0032-numpy-support-in-dynamo.md
albanD: Ho I misread the "not" here I think :D
Maybe replace the end to "to start using PyTorch as a backend for their numpy code in their codebases".

albanD: Also we should update `PyTorch 2.0` -> `torch.compile` :D

lezcano: done and done

lezcano:
The PR upstreaming this work was merged yesterday. Could you please merge this one @albanD ?

PR #57, opened 2023-09-20
[RFC][TorchElastic] topology info in training apps/ranks in

kurman:
Proposal to provide topology information to training apps/ranks that can be implemented as part of TorchElastic.

### DIFF
--- a/RFC-0033-TorchElastic-TopologyInfo.md
+++ b/RFC-0033-TorchElastic-TopologyInfo.md
@@ -1,0 +1,191 @@
+# Providing Topology Information to Training Apps/Ranks
+
+**Authors:**
+* @kurman
+
+
+## Summary
+Provide topology information to each rank/trainer process that can be used by trainers to optimize and potentially automate pytorch distributed models.
+
+
+## Motivation
+As the size of jobs are increasing, ML jobs utilize parallelism to:
+1. increase throughput and 
+2. to allow model sizes that don’t fit into a single GPU or even node. 
+
+Those techniques are known as Data Parallelism (DP), Model Parallelism (MP) and Pipeline Parallelism (PP). Each of them have specific trade-offs and can be combined.
+
+Using those techniques requires some knowledge of underlying topology to make right tradeoffs based on communication overhead (speed, throughput), available GPU memory and use of non-accelerator resources (CPU/RAM). PyTorch has come up with various generic distributed modeling solutions like DTensor and DeviceMesh. However pytorch distributed primitives provide no automation today for taking advantage of topology infor,atopm to perform model partitioning, it is the job of the user to configure the 2-D or 3-D (MP+DP+PP) parallelism solutions in a topology aware manner for optimal performance.
+
+Underlying topology data can inform application on communication constraints that can be used to optimize distributed computation graph that will take communication pattern (however still relies underlying comms to actually use them). Regardles of what are the available communication channel sematics, most value will be is grouping nodes by locality. See more details under [usecases](#Usecases) section.
+
+## Scope
+The proposal focuses on exposing information about infrastructure to application layer for optimization purposes:
+- Contract that applications can use to discover toplogy information when available
+- Initial reference implementation that can be substituted by other implementations
+
+Primary focus on defining contract that can be expanded and other potential approaches/algorithms for defining toplogy information is out of scope.
+
+## Proposed Implementation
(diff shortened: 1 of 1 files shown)


RFC-0033-TorchElastic-TopologyInfo.md
```diff
@@ -0,0 +1,115 @@
...
+
+Eg:
+```json
+{
+  "version": "0.1",
+  "ranks": {
+    "0": {
+      "peers": {
+        "1": {
+          "connection": {
```
d4l3k, line 72: Are we encoding all possible connections between hosts? I don't see a field for type to differentiate network vs IB vs etc

Are we considering latency for loaded vs unloaded? TCP/ethernet has very different latency there whereas IB is more consistent from my understanding

If IB is present I can't really imagine a case where we would care about ethernet

RFC-0033-TorchElastic-TopologyInfo.md
```diff
@@ -0,0 +1,115 @@
...
+            "latency": {
+              "value": "22",
+              "measurement": "us"
+            },
+            "bandwidth": {
+              "value": "64",
+              "measurement": "GB/s"
+            },
+            "channels": {
+                "value": "4"
```
d4l3k, line 82: is this number of connections or lanes?  multi-NIC vs # of nvlink lanes?

d4l3k (commented):
Do we have any details on what will consume `WORLD_TOPOLOGY_DETAILS`? Also can you share any details on how we'll compute the graph? Integrating with things like IB/NVLink seem pretty cluster specific -- can we autodetect that topology in all cases?

Also wondering about things like AWS with spline topologies etc which impact distributed performance pretty significantly

Profiling bandwidth/latency also seems tricky when there are many nodes so would be nice to see some details on that

RFC-0033-TorchElastic-TopologyInfo.md
```diff
@@ -0,0 +1,115 @@
...
+|-----|---------------------------|-------------------------|----------------------|
+|RANK0|              X            | NVLink [22us,64GB/s, 4] | IB[600us, 0.4GB/s,4] |
+|RANK1|NVLink [22us,64GB/s, 4]    |            X            | IB[600us, 0.4GB/s,4] |
+|RANK2|IB[600us, 0.4GB/s,4]       | IB[600us, 0.4GB/s,4]    |           X          |
+
+
+
+Out-of-scope for now:
+- System resources allocated for each rank (eg MEM/CPU/GPU) 
+  - Those tend to be homogeneous
```
kiukchung, line 53: network may not always be homogeneous (the NIC would be homogeneous, but the network route path may not - e.g. if you can allocated among multiple racks/bricks/clusters).

RFC-0033-TorchElastic-TopologyInfo.md
```diff
@@ -0,0 +1,115 @@
...
+  - Most of the can be easily detected at runtime by the trainer code
+- More fine-grained details based on communication pattern (p2p vs collectives)
+
+### **Format of the topology information file**
```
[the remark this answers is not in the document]
kurman: +on API, adding that to proposal.

On additional datasources: we need a single point of discovery of this data at application level therefore it has to be controlled via underlying infra setup. We can extend to other datasources but I believe factory mechanism should be encapsulated.

RFC-0033-TorchElastic-TopologyInfo.md
kurman: clarified, I meant non-topology.

RFC-0033-TorchElastic-TopologyInfo.md
kurman: Thanks, I have added connection type field.

My take is that in some cases we would like to know that there are other connectivity types are available. For example we can utilize eth+CPU for non-accelerator work such as peer checkpointing.

RFC-0033-TorchElastic-TopologyInfo.md
```diff
@@ -0,0 +1,115 @@
...
+    },
+    "1": {"peers": {...}}
+  }
+}
+```
+
+
+### **Implementation details**
+[torchrun/torchelastic](https://pytorch.org/docs/stable/elastic/run.html) will 
+- Build the topology representation during the rendezvous process
```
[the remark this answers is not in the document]
kurman: Good points, tried to address them all in the `defining/discovering topology section`

PR #60, opened 2023-10-21
Proposal for implementation of Lion Optimizer.

RahulVadhyar:
This is a small proposal for the implementation of Lion Optimizer.


albanD:
Hey!
Thanks for sending an rfc. I think this is small enough that we can continue the discussion on the related issue directly: https://github.com/pytorch/pytorch/issues/94904

PR #61, opened 2023-12-24
RFC-0034-batch-kronecker-product

dannygoodacre:


RFC-0034-batch-kronecker-product.md
```diff
@@ -0,0 +1,170 @@
...
+a.shape = (2,5,2,3)
+b.shape = (2,5,6)
+batch_kron(a, b, 0).shape = (2,5,10,18)
+batch_kron(a, b, 1).shape = (2,5,2,18)
+```
+We see that after the batch dimensions, the rest of the dimensions are right-aligned and Kroneckered as usual.
+
+
+## **Drawbacks**
+This is not a breaking change.
```
albanD, line 128: Adding this new kwarg to `torch.kron` wouldn't be bc-breaking either but would make it diverge from numpy's kron.
cc @rgommers any plans on adding kron to the arrray API that we should follow here?

RFC-0034-batch-kronecker-product.md
```diff
@@ -0,0 +1,170 @@
...
+Expand the PyTorch Kronecker product to allow for batching the product along a specified dimension.
+
+
+## **Motivation**
+
+This feature is motivated by a need for such a function in a project I am working on, which simulates the evolution of quantum spin systems. Since this feature can be generalised from matrices to tensors --- as the Kronecker product has already been in PyTorch --- I feel that it belongs within PyTorch itself.
+
+
+## **Proposed Implementation**
+In the following, `a` and `b` are tensors of the specified sizes, and `batch_kron(a, b, batch_dim)` the proposed feature, where `batch_dim = 0` refers to the outermost dimension.
```
albanD, line 42: Not sure what you mean by outermost here. But 0 should refer to the first dimension indeed.
Note that to match existing function, we can also consider doing the same as `mm`/`bmm` and only have the 0th dimension always be batch for a `bkron` op.

RFC-0034-batch-kronecker-product.md
```diff
@@ -0,0 +1,170 @@
...
+b.shape = (3,5,3)
+```
+could only be batched along dimension 0.
+
+Some more examples:
+
+```
+a.shape = (2,5,4)
+b.shape = (2,5,7)
+batch_kron(a, b, 0).shape = (2,25,28)
```
albanD, line 62: cc @zou3519 do you think that using vmap here would be the best way to expose this kind of feature rather than adding a new function? Can we implement a fast batched kron without having a native op for it?

albanD (commented):
Hi,

Thanks a lot for the detailed rfc!
I do think it makes sense to add this feature, I do have a couple inline comments.

RFC-0034-batch-kronecker-product.md
dannygoodacre: Keeping consistent with `bmm` makes sense. Then the only condition on the tensors would be that the first dimensions agree. Plus, batching in any other dimension is rather niche and potentially overkill to include.

PR #63, opened 2024-03-08
RFC-0036-hardware-accelerators-pytorch.org

bsochack:


hipudding (commented):
Thanks to Intel colleagues for the excellent proposal. We fully support it. This proposal not only provides an opportunity for all hardware accelerators integrated with PyTorch to demonstrate their capabilities but also offers PyTorch users more comprehensive installation and usage guidance.

RFC-0036-hardware-accelerators-pytorch.org.md
```diff
@@ -0,0 +1,89 @@
...
+## **Proposed Implementation**
+The proposal is to enhance pytorch.org by adding 3  new sections:
+* Add new compute platforms to “Install PyTorch” section at the main page of pytorch.org.
+* Add new compute platforms to https://pytorch.org/get-started/.
+* Add new “Compute Platforms” section at the main page of pytorch.org.
+
+### PyTorch integration types
+There are at least 2 ways how compute platforms are integrated with PyTorch:
+1. In-tree – CPU, CUDA, ROCm are developed, built and tested in PyTorch environment. PyTorch is ensuring quality criteria. This approach is limited to only a few compute platforms and it does not scale with number of compute platforms.
+2. Out-of-tree – Integration of other compute platforms like Intel Gaudi is done via additional python package (extension) that needs to be installed on top of PyTorch CPU package. Development, built and testing is done outside of PyTorch. In this case:
```
hipudding, line 34: Does "out of tree" hardware accelerator refer to an accelerator that implements logic outside of PyTorch upstream, including devices accessed via privateuse1?

If so, could we specify "out-of-tree (including privateuse keys)" here?

RFC-0036-hardware-accelerators-pytorch.org.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Stable vs. nightly builds
+1. Stable builds – the hardware provider shall provide the installation commands when a compute platform is tested against given PyTorch version and it meets the quality criteria set by PyTorch Foundation.    
+2. Preview (nightly) -  Similar to the stable builds, but the hardware provider must implement method to provide quickly fixes for PyTorch nightly.   It should be optional.
+
+### Quality criteria
+PyTorch foundation shall introduce minimal requirements for new compute platforms.
+Report what kind of testing was done on a compute platform with PyTorch build
+* A common test report format - to be defined.
+  * A test report to contain: results of tests (PyTorch UT, TorchBench, model runs, additional test suites), tested OSes.
```
hipudding, line 46: Due to variations in capabilities and implementation among different hardware accelerators, how should the minimum testing suite be defined?

hipudding:
If it's possible, we would like to apply to become co-authors, aiming to collaboratively promote the implementation of this proposal.

RFC-0036-hardware-accelerators-pytorch.org.md
bsochack: Yes, "out of tree" refers to "integration logic" that is implemented outside of PyTorch upstream. It is applicable to privateuse1 and other accelerators with device specific dispatch key (i.e. Gaudi uses HPU key)

RFC-0036-hardware-accelerators-pytorch.org.md
bsochack: This is a challenge. One option is to verify basic infrastructure and ops, i.e. tensor management, dispatcher, copy ops, basic math, hello world, etc.
The device capabilities reporting in PyTorch is not very extensive and extending it to take care of all hardware variants would be overengineering.

RFC-0036-hardware-accelerators-pytorch.org.md
```diff
@@ -0,0 +1,87 @@
...
+Out-of-tree integration refers to compute platforms integrated with its own device key or PrivateUse1.
+Example of out-of-tree integration with its own device key is Intel Gaudi with "HPU" key.
+Example of out-of-tree integration using PrivateUse1 is Huawei Ascend.
+
+
+### Stable vs. nightly builds
+1. Stable builds – the hardware provider shall provide the installation commands when a compute platform is tested against given PyTorch version and it meets the quality criteria set by PyTorch Foundation.    
+2. Preview (nightly) -  Similar to the stable builds, but the hardware provider must implement method to provide quickly fixes for PyTorch nightly.   It should be optional.
+
+### Quality criteria
```
albanD, line 45: I think we want to go a bit deeper into the details of what are going to be the criterions and expectations here.
I am also working on this and I think we need to properly define a couple of "tiers" for backends, defining expectations in details. I think that we can then refine the way we present them from this proposal where top-tier backends will be in the main get-started page, all others will be in the Compute Platforms page linked from get-started.

I'll take a stab at that proposal and come back to you here.

albanD (commented):
Hey!

Thank you very much for this proposal.
It all makes a lot of sense.
In particular:
- Adding the Accelerators dropdown looks great
- Adding a "Compute Platforms" landing page with sub-page for each system sounds good as well. The website is being re-organized right now, we most likely want to wait for this to be done to see what is the best place to put it.
- I don't think we need to make too much distinction between in-tree and out-of-tree backends. Our goal is for the experience between the two to be the same and we need different install commands for all backends anyways.
- The "Quality Criteria" section is the one that we need to clarify the most. Leaving comments inline.

RFC-0036-hardware-accelerators-pytorch.org.md
bsochack: @albanD waiting for the proposal from your side.
Please let me know if I can help here.

thiagocrepaldi:
FYI @wschin @xadupre

RFC-0036-hardware-accelerators-pytorch.org.md
bsochack: My high level thoughts about quality criteria:

1. There are many types of devices and there should be a way for them to pass quality criteria
- some of them are specialized i.e. inference only, machine learning or more general purpose like GPUs or CPUs
- they differ in capabilities i.e. data types support, distributed ops support
- there should be a way to mark relevant tests for a given device through test marking or more extensive capability reporting by a device
2. PyTorch has lots of unit tests, but they are very often specialized to CUDA
- it will take a few PyTorch releases to generalize tests for other devices i.e. we have number of generic changes that will enable Intel Gaudi and other devices with PT UTs. We will follow up with RFCs and UT PRs 
- PyTorch UTs are probably the best ones to verify quality of PyTorch support by a device
3. The quality criteria should be set to reasonable level until PyTorch UT changes are upstreamed i.e. initially start with 50% pass rate for PT 2.4 and gradually increase it in the next releases
4. I agree that there should be a couple of "tiers"
- Do you expect that 1st tier out of tree backends (the ones from the main get-started page) will work with PyTorch CPU from pytorch.org?

bsochack: @albanD any update?

bsochack: @albanD are there any details that can be shared with us?

albanD: Hey!
Sorry I have been held out in other projects (PT conference right now) and was not able to look at this yet.

AnthonyBarbier: This thread seems to have gone stale but IMO this is a crucial part of this proposal: the tests need to be refactored to be gated on features sets rather than specific backends or technology. (i.e no `if cuda` or `if device == "cpu"`)

You could then use the features enabled for a given backend as part of your reporting: "This platform is stable and supports torch.compile with a Triton backend for the following dtypes. Note: non-contiguous tensors are not supported", etc.

PR #67, opened 2024-08-27
Update dead links with permalinks and archived versions

hvaara:
RFCs are great artifacts for new people (like myself) who want to learn about certain aspects of the projects.

Unfortunately line-number anchors tend to drift and content gets moved, resulting in dead links. I've updated most of the published RFCs with permalinks and archived versions for the links that resulted in 404s or confusing anchor points.


albanD (approved):
Thanks for taking the time to fix this up!

PR #69, opened 2024-09-27
RFC-0001-economic-dataloader.md

yoadbs:
A new dataloader multiprocessing pipeline design is suggested. This pipeline splits the task of batch generation, into 2 types of workers: item generating workers, and batch generating workers. This pipeline is designated to significantly reduce random-access-memory (RAM) usage, without any significant reduction in throughput.


albanD:
cc @andrewkho

andrewkho:
Hi @yoadbs thank you for this thoughtful RFC! I just had a quick look but this looks like it would be covered by some of our plans in torchdata to allow more modular parallelism: https://github.com/pytorch/data/issues/1318 . I know it's long but I believe it should cover your use case as well, please let me know if it doesn't.

Some thoughts on this in general, these will be true for both your RFC and the one in pytorchd/data's #1318:
+ this is going to be more relevant for large batch_size and larger data, eg HD video
- Introducing more IPC between worker pools might slow things down
- requires tuning multiple worker pools/prefetch buffers

PR #73, opened 2025-05-13
RFC for vector length agnostic SVE Vectorized class

Ryo-not-rio:
This is a proposal for a Vectorized class for SVE to replace the current Vectorized class which will make it vector length agnostic, avoiding code duplication in the future

### DIFF
--- a/RFC-0044-sve-vectorized-class.md
+++ b/RFC-0044-sve-vectorized-class.md
@@ -64,11 +64,14 @@ class Vectorized<float> {
 ### Implementation cost
 This is a large change which requires an overhaul of all of the current SVE `Vectorized` as well as any code that expects the size() function to be constexpr. The first cost can be mitigated by updating the `Vectorized` classes one by one.
 
-### Sideffects from non-constexpr size()
+### Side-effects from non-constexpr size()
 By changing the size() to non-constexpr, we will be changing a large part of the codebase which may cause regressions. These will need to be benchmarked thoroughly and if we choose to accept any regressions, they will need to be limited to aarch64 architectures.
 
 ### Memory footprint increase
 By storing an array with the size "max SVE vector length (2048 bits being the maximum possible and 512 bits being the longest hardware available)", the memory footprint is increased by `2048 bits x number of existing Vectorized classes`. Since `Vectorized` classes are created and destoryed in loops with only a few instances existing simultaneously, we expect this effect to be minimal, but we should benchmark this using actual models. We could also limit this effect by using the maximum vector size currently available on hardware with scope to change this if necessary.
+
+### Side-effects from using -O3
+Using `-O3` increases the package size of PyTorch, which is not an issue for use cases using the PyPI package where we already use `-O3` or for devices where package size is not a large concern such as infrastructure. However, it becomes an issue for clients such as mobile where package size may matter. In these instances where PyTorch is built using lower optimizaion flags, there will be a regression compared to the current implementation.
 
 ## **Benchmarking plan**
 To mitigate the risk from changing the size() from constexpr to const, we propose the following order of patches to PyTorch:
--- a/RFC-0044-sve-vectorized-class.md
+++ b/RFC-0044-sve-vectorized-class.md
@@ -1,0 +1,112 @@
+# [Vector length agnostic SVE class]
+
+**Authors:**
+* @Ryo-not-rio
+
+
+## **Summary**
+A vector length agnostic implementation of the `Vectorized` class for SVE.
+
+## **Motivation**
+PyTorch contains a "Vectorized" class that wraps different SIMD intrinsics for different architectures by having a vector as a class attribute, and the class methods acting as the wrapper around the intrinsics.  This works well for non-scalable vectors but poses an issue for SVE due to the inability to store them as class attributes. The current workaround for this is to use the compiler flag -msve-vector-bits=\<bits\>, however this is not ideal as this would 1. require separate vectorized classes for different vector lengths and 2. does not allow for runtime detection of the actual vector length. We currently only have an implementation of the Vectorized class for 256-bit SVE machines but as we think about adding support for different vector length, we need to consider how to avoid code duplication as raised by @malfet [here](https://github.com/pytorch/pytorch/pull/138388#issuecomment-2635612409). This RFC aims to solve the issue but creating a Vectorized class that detects the vector length at runtime as SVE is intended to be used, allowing us to support different vector lengths without writing any duplicate code.
+
+## **Proposed Implementation**
+The basic premise of our proposal is to store not the vector but an array in our vectorized class which we will load from and store to with each operation. A minimal version is shown at the end.
+
+Now this introduces quite an obvious overhead of an additional load and store operation with each op. However, the compiler is able to optimize these out with the following conditions:
+
+1. The -O3 flag is set
+2. The `svptrue_b32()` predicate is used
+3. You are storing to and then loading from the same pointer
+
+Ensuring these conditions are met and by inlining the functions, we can rely on the compiler to optimize the duplicate load and stores, ensuring we do not introduce any regressions.
+
+### The size problem
+We face a challenge with this implementation due to the constraint of the size() function being constexpr. The size() function which returns the number of elements in the Vectorized class cannot be constexpr in our implmentation due to SVE vector lengths being unknown at compile time. We propose we change this to be const instead of constexpr.
+
+```
+class Vectorized<float> {
+	float values[64]; // Maximum number of elements supported by any SVE machine
+
+    static inline const size_type size() {
+        return svcntw();
+    }
+    static inline Vectorized<float> loadu(const float * vs) {
+        Vectorized<float> v;
+        svfloat32_t vec = svld1_f32(svptrue_b32(), static_cast<const float *>(vs));
+        svst1_f32(svptrue_b32(), v.values, vec);
+        return v;
+    }
+
+    inline void store(void* ptr) const {
+        svfloat32_t vec = svld1_f32(svptrue_b32(), values);
+        svst1_f32(svptrue_b32(), static_cast<float *>(ptr), vec);
+    }
+
+    inline Vectorized<float> abs() const {
+		svfloat32_t v = svld1_f32(svptrue_b32(), values);
+    	v = svabs_f32_x(svptrue_b32(), *this);
+     	svst1_f32(svptrue_b32(), values, v);
+		return *this;
+  	}
+}
+```
+
+## **Metrics **
+- Reduction of code duplication
+- Speedup of PyTorch on SVE machines with non-256 bit vectors
+  - Softmax sped up by 2.73x on Neoverse V2
+  - X * sigmoid * softmax sped up by 1.65x on Neoverse V2
+- No speed or accuracy regression on 256-bit vectors
+
+
+## **Drawbacks**
+### Implementation cost
+This is a large change which requires an overhaul of all of the current SVE Vectorized as well as any code that expects the size() function to be constexpr. The first cost can be mitigated by updating the Vectorized classes one by one, but the size() change will need to be done all at once.
+
+### Sideffects from non-constexpr size()
+There are a number of functions that use the size() function to initialize an array. These will have to be changed to an alternative such as a vector. Since a vector is implemented as an array under the hood, we hope this will not cause any regressions but a thorough benchmarking of these functions need to be done to ensure that this is the case.
+
+## **Alternatives**
+To keep the size() function constexpr, we considered setting the size of the Vectorized class to be the maximum possible SVE vector length (currently 512 bits) and loading multiple vectors as necessary. However, this poses the following problems:
+
+1. Increases the tail loop size unecessarily for machines with smaller vector lengths.
+2. Compiler can no longer optimize out duplicate loads and stores as multiple vectors need to be handled consecutively.
+3. Detection of number of vectors needed introduces an extra overhead
+
+Due to these issues combined, especially 2., this alternative introduces a ~30% overhead compared to the current implementation.
+
+## **Unresolved questions**
+* How much of the existing code has to change due to changing the size() from constexpr to const
+* What if any performance regression will we see due to changing size() from constexpr to const?
+* Will we have to potentially have to write code specific to SVE due to the size() change?
+
+
+## Resolution
+TBD
+
+### Level of Support
+Choose one of the following:
+* 1: Overwhelming positive feedback.
+* 2: Positive feedback.
+* 3: Majority Acceptance, with conflicting Feedback.
+* 4: Acceptance, with Little Feedback.
+* 5: Unclear Resolution.
+* 6: RFC Rejected.
+* 7: RFC Rejected, with Conflicting Feedback.
+
+
+#### Additional Context
+TBD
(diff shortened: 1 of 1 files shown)


RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,112 @@
...
+
+## **Drawbacks**
+### Implementation cost
+This is a large change which requires an overhaul of all of the current SVE Vectorized as well as any code that expects the size() function to be constexpr. The first cost can be mitigated by updating the Vectorized classes one by one, but the size() change will need to be done all at once.
+
+### Sideffects from non-constexpr size()
+There are a number of functions that use the size() function to initialize an array. These will have to be changed to an alternative such as a vector. Since a vector is implemented as an array under the hood, we hope this will not cause any regressions but a thorough benchmarking of these functions need to be done to ensure that this is the case.
+
+## **Alternatives**
+To keep the size() function constexpr, we considered setting the size of the Vectorized class to be the maximum possible SVE vector length (currently 512 bits) and loading multiple vectors as necessary. However, this poses the following problems:
```
jondea: Might be worth clarifying that 2048 is the maximum possible, 512 is the maximum available hardware

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,112 @@
...
+Now this introduces quite an obvious overhead of an additional load and store operation with each op. However, the compiler is able to optimize these out with the following conditions:
+
+1. The -O3 flag is set
+2. The `svptrue_b32()` predicate is used
+3. You are storing to and then loading from the same pointer
+
+Ensuring these conditions are met and by inlining the functions, we can rely on the compiler to optimize the duplicate load and stores, ensuring we do not introduce any regressions.
+
+### The size problem
+We face a challenge with this implementation due to the constraint of the size() function being constexpr. The size() function which returns the number of elements in the Vectorized class cannot be constexpr in our implmentation due to SVE vector lengths being unknown at compile time. We propose we change this to be const instead of constexpr.
```
jondea: It may be worth noting that I think a const virtual function can be overridden by a constexpr function. And even if not, I suspect in many cases the compiler should be able to work it out if the class is concrete

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,112 @@
...
+
+## **Proposed Implementation**
+The basic premise of our proposal is to store not the vector but an array in our vectorized class which we will load from and store to with each operation. A minimal version is shown at the end.
```
jondea: I assume here you don't mean std::vector, might be worth clarifying.

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,112 @@
...
+3. You are storing to and then loading from the same pointer
+
+Ensuring these conditions are met and by inlining the functions, we can rely on the compiler to optimize the duplicate load and stores, ensuring we do not introduce any regressions.
+
+### The size problem
+We face a challenge with this implementation due to the constraint of the size() function being constexpr. The size() function which returns the number of elements in the Vectorized class cannot be constexpr in our implmentation due to SVE vector lengths being unknown at compile time. We propose we change this to be const instead of constexpr.
+
+```
+class Vectorized<float> {
+	float values[64]; // Maximum number of elements supported by any SVE machine
```
jondea, line 29: What's the expected additional memory/allocation overhead of having the max? I assume that there are relatively few of the these classes around and they are not created/destroyed in hot loops, so it should be minimal.

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,112 @@
...
+
+## **Motivation**
+PyTorch contains a "Vectorized" class that wraps different SIMD intrinsics for different architectures by having a vector as a class attribute, and the class methods acting as the wrapper around the intrinsics.  This works well for non-scalable vectors but poses an issue for SVE due to the inability to store them as class attributes. The current workaround for this is to use the compiler flag -msve-vector-bits=\<bits\>, however this is not ideal as this would 1. require separate vectorized classes for different vector lengths and 2. does not allow for runtime detection of the actual vector length. We currently only have an implementation of the Vectorized class for 256-bit SVE machines but as we think about adding support for different vector length, we need to consider how to avoid code duplication as raised by @malfet [here](https://github.com/pytorch/pytorch/pull/138388#issuecomment-2635612409). This RFC aims to solve the issue but creating a Vectorized class that detects the vector length at runtime as SVE is intended to be used, allowing us to support different vector lengths without writing any duplicate code.
```
jondea: nit: you refer to the `Vectorized` class in 4 different ways (backticks, speechmarks, caps etc)

jondea:
I left a few minor comments, but I think overall it looks promising! Thank you

albanD (commented):
cc @malfet @swolchok @kimishpatel @guangy10 as people working on this topic in case you have any comments

kimishpatel:
> cc @malfet @swolchok @kimishpatel @guangy10 as people working on this topic in case you have any comments

thanks Alban. Will take a look tomorrow or friday

RFC-0044-sve-vectorized-class.md
swolchok: I don't think there are virtual functions involved; architectures specialize Vectorized<T>

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,112 @@
...
+  - X * sigmoid * softmax sped up by 1.65x on Neoverse V2
+- No speed or accuracy regression on 256-bit vectors
+
+
+## **Drawbacks**
+### Implementation cost
+This is a large change which requires an overhaul of all of the current SVE Vectorized as well as any code that expects the size() function to be constexpr. The first cost can be mitigated by updating the Vectorized classes one by one, but the size() change will need to be done all at once.
+
+### Sideffects from non-constexpr size()
+There are a number of functions that use the size() function to initialize an array. These will have to be changed to an alternative such as a vector. Since a vector is implemented as an array under the hood, we hope this will not cause any regressions but a thorough benchmarking of these functions need to be done to ensure that this is the case.
```
swolchok: I would be concerned about this. Arrays go on the stack; vectors go on the heap.

RFC-0044-sve-vectorized-class.md
swolchok: Is "size() determines the size of an array" the only use of constexpr size()? If so, we should have a few options to solve that problem:
- rely on the VLA extension (I'm mobile so I can't look up which compilers do and don't support this, but I'd expect clang and gcc to get it right and msvc to be annoying) to size arrays this way anyway; we would just have to use C arrays instead of std::array
- use some template metaprogramming magic to create a container templated on vectorized that detects whether size is constexpr. Create an array if so, and either a padded-to-max-length-for-sve array or a vector otherwise.

swolchok: Some quick research says we can generally detect whether a function is constexpr when we get C++20, which should be soon: https://quuxplusone.github.io/blog/2022/01/04/test-constexpr-friendliness/

However, since this is a very specific case, we could just create a custom vectorized_of_t_has_constexpr_size traits class that defaults to true and is specialized for VLA-SVE-specialized Vectorized<T> types to say false.

Ryo-not-rio: Another option to keep size() constexpr is to use a hybrid approach where we decide the size() on compile time. This will keep size() constexpr but will not be able to take advantage of SVE's runtime vector length detection feature

Ryo-not-rio: After further investigation, changing size() to const would require a rewrite of the VectorizedN class as well so I'm leaning towards the hybrid approach mentioned above

RFC-0044-sve-vectorized-class.md
Ryo-not-rio: Added information on this!

RFC-0044-sve-vectorized-class.md
Ryo-not-rio: fixed

RFC-0044-sve-vectorized-class.md
Ryo-not-rio: fixed

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,115 @@
...
+## **Proposed Implementation**
+The basic premise of our proposal is to store not the SVE vector but an array in our `Vectorized` class which we will load from and store to with each operation. A minimal version is shown at the end.
+
+Now this introduces quite an obvious overhead of an additional load and store operation with each op. However, the compiler is able to optimize these out with the following conditions:
+
+1. The -O3 flag is set
+2. The `svptrue_b32()` predicate is used
+3. You are storing to and then loading from the same pointer
+
+Ensuring these conditions are met and by inlining the functions, we can rely on the compiler to optimize the duplicate load and stores, ensuring we do not introduce any regressions.
```
cfRod, line 22: Is there any risk of different behaviour across different compilers/versions?

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,115 @@
...
+     	svst1_f32(svptrue_b32(), values, v);
+		return *this;
+  	}
+}
+```
+
+## **Metrics **
+- Reduction of code duplication
+- Speedup of PyTorch on SVE machines with non-256 bit vectors
+  - Softmax sped up by 2.73x on Neoverse V2
```
cfRod, line 58: Are these numbers similar for eager mode or compile?

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,115 @@
...
+## **Alternatives**
+To keep the size() function constexpr, we considered setting the size of the `Vectorized` class to be the maximum possible SVE vector length and loading multiple vectors as necessary. However, this poses the following problems:
+
+1. Increases the tail loop size unecessarily for machines with smaller vector lengths.
+2. Compiler can no longer optimize out duplicate loads and stores as multiple vectors need to be handled consecutively.
+3. Detection of number of vectors needed introduces an extra overhead
+
+Due to these issues combined, especially 2., this alternative introduces a ~30% overhead compared to the current implementation.
+
+## **Unresolved questions**
```
cfRod, line 93: Are there any plans to add specific tests?

Ryo-not-rio: For unit tests, the existing tests should cover us, for perf tests, we should run the whole set of integration tests

RFC-0044-sve-vectorized-class.md
Ryo-not-rio: For eager mode

Ryo-not-rio:
@swolchok I've scoped out the all the changes that need to be done to change `size()` to not const-expr and added a benchmark plan to make these changes, please have a look and let me know what you think

RFC-0044-sve-vectorized-class.md
Ryo-not-rio: Yes, I guess we'll have to make sure this optimization happens on all our supported compilers

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,123 @@
...
+
+## **Proposed Implementation**
+The basic premise of our proposal is to store not the SVE vector but an array in our `Vectorized` class which we will load from and store to with each operation. A minimal version is shown at the end.
+
+Now this introduces quite an obvious overhead of an additional load and store operation with each op. However, the compiler is able to optimize these out with the following conditions:
+
+1. The -O3 flag is set
```
kimishpatel, line 18: -O3 can introduce other issues, including code bloat. how do you handle that?

RFC-0044-sve-vectorized-class.md
```diff
@@ -0,0 +1,123 @@
...
+     	svst1_f32(svptrue_b32(), values, v);
+		return *this;
+  	}
+}
+```
+
+## **Metrics **
+- Reduction of code duplication
+- Speedup of PyTorch on SVE machines with non-256 bit vectors
+  - Softmax sped up by 2.73x on Neoverse V2
```
kimishpatel, line 58: I think one question i have is this: true power of SVE I presume is that you can have same binary compiled for one implementation of sve work on another and realize performance gain. That is great. If this means that pytorch distribution built for arm, when pip installed, can take advantage of vector length of that machine. is that how we plan to realize performance gain?

RFC-0044-sve-vectorized-class.md
Ryo-not-rio: If I'm not mistaken, PyTorch is built with -O3 in release mode so there shouldn't be any extra code bloat

RFC-0044-sve-vectorized-class.md
Ryo-not-rio: Yes, currently, we only have SVE support for 256bit vector machines, with this change we can do as you suggested: build one version of PyTorch for aarch64 and use it on all aarch64 machines without worrying about vector length

RFC-0044-sve-vectorized-class.md
kimishpatel: That for pypi release, but deployment of pytorch may get built with different flags. Plus note that this being header only change also allows reuse of the class in other implementations including that of custom ops that dont live inside pytorch. And my custom op lib may not be built with O3 for size reasons. Imagine for example deploying this on mobile. so I do think we should touch upon this aspect carefully and at least call it out clearly.

swolchok:
> @swolchok I've scoped out the all the changes that need to be done to change `size()` to not const-expr and added a benchmark plan to make these changes, please have a look and let me know what you think

this doesn't seem to address my concerns about adding additional heap allocations. are there technical barriers to the plan I outlined to avoid heap allocations, such as reasons for size() to be constexpr other than determining the size of an array? We don't need benchmarking to tell us that heap allocations are worse than no heap allocations.

Ryo-not-rio:
>are there technical barriers to the plan I outlined to avoid heap allocations, such as reasons for size() to be constexpr other than determining the size of an array

So I've investigated all the places where size() needs to be constexpr [here](https://github.com/Ryo-not-rio/pytorch/commit/fa05c1de3340215da5dc0a32612e75e2816fc143) and as you suggest, a lot of it just initializing an array which can be solved by using c arrays. The other places where it's used are template parameters where we would have to move the parameter from the template to a function argument

RFC-0044-sve-vectorized-class.md
Ryo-not-rio: Thanks for your insight, that is something I hadn't considered so I'll put it into the RFC. I'm not sure if there's any mitigation we could do at the moment, I'll try less aggressive optimization flags to see how much the compiler can optimize this aspect

malfet:
I probably commented on PyTorch issue rather than RFC, so repeating here this [comment](https://github.com/pytorch/pytorch/issues/153471#issuecomment-2880442090):
> constexpr size allows for a numerous optimizations on the compiler side, and making it dynamic will likely result in a significant slowdown. To accept change like that, that will affect all supported CPU architectures a rigorous test plan is necessary as well as list of HW one must run it against.

Ryo-not-rio:
Since we're adding SVE128 support, I think we can pause this for now, and revisit in the future if needed

PR #74, opened 2025-05-17
Cuda update policy and guide

atalman:
This PR introduces CUDA update policy followed by PyTorch team and consolidates CUDA update guilde from pytorch/builder repo:
https://github.com/pytorch/builder/blob/main/CUDA_UPGRADE_GUIDE.MD

Related RFCs for CUDA update:
https://github.com/pytorch/pytorch/issues/147383
https://github.com/pytorch/pytorch/issues/145544
https://github.com/pytorch/pytorch/issues/138609
https://github.com/pytorch/pytorch/issues/134015
https://github.com/pytorch/pytorch/issues/123456

### DIFF
--- a/RFC-0039-cuda-support.md
+++ b/RFC-0039-cuda-support.md
@@ -1,0 +1,106 @@
+
+# [CPython version support]
+
+**Authors:**
+* @albanD
+
+## **Motivation**
+
+CPython follows annual release cycle. Given that the amount of work to enable and sunset versions yearly is fixed, we should be proactive in how we handle these tasks.
+This RFC suggests an updated policy we want for CPython version support, CI/CD requirements and the proposed yearly timeline for enablement/sunsetting.
+
+The key requirements driving this proposal are:
+- Enable new versions as soon as possible
+- Sunset old versions in a timely manner
+- Set clear long-term expectations for our users
+- Mitigate the risk of a CPython change that would render PyTorch non-functional making it to the CPython final release
+- Minimize any work that would not have been needed if the enablement was done post-final release
+
+
+## **Proposed Implementation**
+
+### Proposed timeline
+
+![Summary timeline](./RFC-0038-assets/cpython_support.png)
+
+### Which CPython version are supported
+
+PyTorch supports all CPython versions that are fully released and have not reached end of life: https://devguide.python.org/versions/
+
+Note: This is an update from the current policy at https://github.com/pytorch/pytorch/blob/main/RELEASE.md#python which is following NumPy’s approach. In practice, we are not following the rules stated in that .md and following the rule stated just above here. We are updating the rule (instead of enforcing the NEP above) as the cost of supporting more versions is minimal.
+Also split build to have a single (large) c++ package shared by all CPython version is way underway (https://github.com/pytorch/pytorch/pull/129011) and will reduce the binary size use on PyPi even though we support a large number of versions.
+
+### What is tested in CI/CD
+
+The goal here is for us to ensure coverage of testing while reducing the cost on CI.
+
+At any given time, we should run:
+- General CI on PR and trunk should run on the oldest supported version.
+- Maintainers can ask for specific CI shard to run on specific versions:
+- Long term for testing features tightly bound to CPython versions (for example Dynamo).
+- Temporarily for enablement work (for example while a new CPython version is being enabled).
+- CD for docker and binaries should run for all supported versions.
+- Wheel/release smoke test should run on all supported versions.
+
+
+### Detailed CPython new version enablement
+
+- Enable new version basic compilation
+  - When: Once the first beta version is released (Usually in May)
+  - Planned for 3.13: 5/07
+  - ETA: 1-2 weeks. Before beta 2
+  - Goal: Fix compilation issues allowing to compile PyTorch locally. Report any needed patch in CPython before beta 2.
+  - Who: Core Maintainers for pytorch/pytorch + relevant maintainers for managed domains
+  - Note: If a patch is needed in CPython, it will be made available easily until next beta release happens
+- Enable new version CI/CD Infrastructure
+  - When: As soon as basic compilation is done
+  - ETA: 2-3 weeks. 
+  - Goal: Generate nightly wheels and scaffolding for new version testing
+  - Who: Core Infra
+- High risk enablement work for the new version  for submodules
+  - When: as soon as the CI/CD infrastructure is done
+  - ETA: before CPython first RC
+  - Planned for 3.13: 07/30
+  - Goal: Verify high risk systems and report any needed patch in CPython such that they can be fixed before RC1
+  - Who: High risk systems owners. As of today this is Dynamo C++ code and python binding subsystems.
+- Low risk enablement work for the new version  for submodules
+  - When: as soon as the CI/CD infrastructure is done
+  - ETA: before CPython final RC
+  - Planned for 3.13: 09/03
+  - Goal: Enable all testing for the new CPython version
+  - Who: Core Maintainers handle the long tail of issues. Specific submodule owner handle larger migration efforts (dynamo, TorchScript, etc)
+- Full new version wheel release
+  - When: When the new cpython verion is officially released (Usually in October)
+  - Planned for 3.13: 10/01
+  - ETA: 2 week
+  - Goal: Update nightlies to track the final release and advertise it as fully working
+  - Who: Infra + Comms
+- Full new version conda release
+  - When: When the new cpython version and our runtime dependencies are supported onconda
+  - ETA: 2 week
+  - Goal: Push binaries that are officially supported
+  - Who: Infra + Comms
+
+
+### Detailed CPython old version Sunset
+
+- Deprecate old version
+  - When: During the release process of the last PyTorch release before the oldest version of python goes EOL
+  - Planned for 3.8: October
+  - Goal: announce deprecation of the oldest cpython version on dev-discuss + release
+  - Who: Comms
+- Stop releasing nightly binaries for oldest version
+  - When: Once the last supported release happened
+  - ETA: 2 weeks
+  - Goal: remove all wheel and conda binaries for the EOL version
+  - Who: Infra
+- Upgrade oldest version CI
+  - When: After nightlies are not published anymore
+  - ETA: 2 weeks
+  - Goal: migrate all the CI jobs running on the oldest version to the soon-to-be oldest version
+  - Who: Infra
+- Remove un-used code and use new features
+  - When: After the version is fully dropped
+  - ETA: N/A
+  - Goal: clean up tech debt related to older cpython versions and simplify code using new features.
+  - Who: All
--- a/RFC-0039-cuda-support.md
+++ b/RFC-0039-cuda-support.md
@@ -72,14 +72,14 @@ So you wanna upgrade PyTorch to support a new CUDA? Follow these steps in order!
 ## 0. Before starting upgrade process
 
 ### A. Currently Supported CUDA and CUDNN Libraries
-Here is the supported matrix for CUDA and CUDNN (versions can be looked up in https://github.com/pytorch/pytorch/blob/main/.ci/docker/common/install_cudnn.sh)
+Here is the supported matrix for CUDA and CUDNN (versions can be looked up in [here](https://github.com/pytorch/pytorch/blob/main/.ci/docker/common/install_cudnn.sh) . Architecture support can be found [here](https://github.com/pytorch/pytorch/blob/main/.ci/manywheel/build_cuda.sh))
 
-| CUDA | CUDNN | additional details |
-| --- | --- | --- |
-| 11.8.0 | 9.1.0.70 | Legacy CUDA Release |
-| 12.6.3 | 9.5.1.17 | Stable CUDA Release |
-| 12.8.0 | 9.7.1.26 | Latest CUDA Release |
-|        | 9.8.0.87 | Latest CUDA Nightly |
+| CUDA | CUDNN | architectures supported | additional details |
+| --- | --- | --- | --- |
+| 11.8.0 | 9.1.0.70 | Kepler(3.7), Maxwell(5.0), Pascal(6.0), Volta(7.0), Turing(7.5), Ampere(8.0, 8.6), Hopper(9.0) | Legacy CUDA Release |
+| 12.6.3 | 9.5.1.17 | Maxwell(5.0), Pascal(6.0), Volta(7.0), Turing(7.5), Ampere(8.0, 8.6), Hopper(9.0)  | Stable CUDA Release |
+| 12.8.0 | 9.7.1.26 | Turing(7.5), Ampere(8.0, 8.6), Hopper(9.0), Blackwell(10.0, 12.0+PTX)  | Latest CUDA Release |
+|        | 9.8.0.87 | Turing(7.5), Ampere(8.0, 8.6), Hopper(9.0), Blackwell(10.0, 12.0+PTX)  | Latest CUDA Nightly |
 
 ### B. Check the package availability
 
--- a/RFC-0039-cuda-support.md
+++ b/RFC-0039-cuda-support.md
@@ -15,7 +15,7 @@ The proposal is to provide two main benefits
 
 ### We should introduce new version of CUDA when
 
-- It enables new important GPU architecture (For example A100 with CUDA-11.1)
+- It enables new important GPU architecture (For example Blackwell with CUDA-12.8)
 - It brings significant performance improvements
 - It fixes significant correctness issues
 - It significantly reduces binary/memory footprint
@@ -31,14 +31,14 @@ As soon as we introduce a new Experimental Version we should consider moving the
 ### Detailed Process of Introducing new CUDA version
 
 1. Evaluate CUDA update necessity
-  Goal: When Any of this is true:
-  - It enables new important GPU architecture (For example A100 with CUDA-11.1)
+  Goal: When any of this is true:
+  - It enables new important GPU architecture (For example Blackwell with CUDA-12.8)
   - It brings significant performance improvements
   - It fixes significant correctness issues
   - It significantly reduces binary/memory footprint
 
 2. Evaluate if we have all packages for update
-  When: As soon as Update determined to be necessary
+  When: As soon as Update determined to be necessary. Start by creating RFC [issue](https://github.com/pytorch/pytorch/issues/145544) with possible CUDA matrix to support for next release.
   Goal: Make sure everything is available to perform complete upgrade of CUDA and dependencies
 
 3. Update CUDA in CD (this is necessary condition to qualify for CUDA version be released as Experimental)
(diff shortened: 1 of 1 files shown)


RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,196 @@
...
+* @atalman @malfet
+
+## **Motivation**
+
+The proposal is to provide two main benefits
+
+- Aligned decision policy provides transparent decision making;
+- Incorporate decision points in the early release process to stagger CUDA migration from the RC release integration work.
+
+## **Proposed Policy and Process**
```
ZainRizvi, line 15: Since this is an existing policy, in the past year when has this resulted in us deciding to _not_ do a cuda upgrade?

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,196 @@
...
+As soon as we introduce a new Experimental Version we should consider moving the previous Experimental Version to Stable, and decommission the previous Stable version. Typically we want to support 2-3 versions of CUDA as follows:
+
+- Optional Legacy Version: If we need to have 1 version for backend compatibility or to work around the current limitation. For example: CUDA older driver is incompatible with newer CUDA version
+- Stable Version: This is a stable CUDA version that is used most of the time. This is the version we want to upload to PyPI. Please note that if the stable version is equal or slightly older to the version then the version fbcode is using, we probably need to keep it in CI/CD system as Optional Legacy Version.
+- Latest Experimental Version: This is the latest version of CUDA that we want to support. Minimal requirement is to have it available in nightly releases (CD)
+
+### Detailed Process of Introducing new CUDA version
+
+1. Evaluate CUDA update necessity
+  Goal: When Any of this is true:
```
ZainRizvi:
```suggestion
  Goal: When any of this is true:
```

ZainRizvi: Are there examples of cuda version updates that do _not_ meet this criteria?

The hard decisions are going to be when someone wants to argue that a upgrade does NOT meet the bar, so having real-world examples of those will help guide those future discussions

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,196 @@
...
+5. Run benchmarks to compare new experimental CUDA version to current stable CUDA version
+  When: Update CUDA in CD  is complete
+  Goal: Make sure we identify all performance regressions
+
+
+### Detailed Process of Deprecating CUDA version
+
+1. Evaluate deprecation of legacy CUDA version from CI/CD
+When: We completed CUDA update and previous experimental CUDA version is qualified to be stable and we have 3 supported versions (legacy, stable and experimental)
+Goal: Make sure we support 2 versions of CUDA, supporting 3 versions can be an exception for certain release where we need to keep legacy version
```
ZainRizvi, line 58: > supporting 3 versions can be an exception for certain release where we need to keep legacy version

Can you share examples of when we have needed to keep a legacy version around in the past?  Sounds like we might incur a significant CI cost to test 50% more CUDA versions

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,196 @@
...
+
+2. Deprecate legacy CUDA version from CI/CD
+When: Evaluate deprecation of legacy CUDA version from CI/CD is complete
+Goal: Support for legacy CUDA versions is dropped, starting from PyTorch Domain Libraries and then in PyTorch core. First we drop CD support and then CI support.
```
ZainRizvi, line 62: > First we drop CD support and then CI support.

Is the order important?

RFC-0039-cuda-support.md
atalman: Yes this is current situation with CUDA 11.8: https://github.com/pytorch/pytorch/issues/147383

RFC-0039-cuda-support.md
atalman: Yes, dropping CD support is easy to do. When Dropping CI support is most likely a migration of these CI jobs to next CUDA version and may take somewhat longer.

RFC-0039-cuda-support.md
atalman: Yes, I believe the history for the updates are here:
https://github.com/pytorch/pytorch/issues/147383
https://github.com/pytorch/pytorch/issues/145544
https://github.com/pytorch/pytorch/issues/138609
https://github.com/pytorch/pytorch/issues/134015
https://github.com/pytorch/pytorch/issues/123456

RFC-0039-cuda-support.md
atalman: Yes, this is a history of decisions we took for past year https://github.com/pytorch/rfcs/pull/74#discussion_r2096005474 

We supported CUDA: 11.8, 12.1, 12.4, 12.6 and 12.8 .

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+- It brings significant performance improvements
+- It fixes significant correctness issues
+- It significantly reduces binary/memory footprint
+
+### We would deprecate version of CUDA when
+
+As soon as we introduce a new Experimental Version we should consider moving the previous Experimental Version to Stable, and decommission the previous Stable version. Typically we want to support 2-3 versions of CUDA as follows:
+
+- Optional Legacy Version: If we need to have 1 version for backend compatibility or to work around the current limitation. For example: CUDA older driver is incompatible with newer CUDA version
+- Stable Version: This is a stable CUDA version that is used most of the time. This is the version we want to upload to PyPI. Please note that if the stable version is equal or slightly older to the version then the version fbcode is using, we probably need to keep it in CI/CD system as Optional Legacy Version.
```
albanD: I don't see why we should talk about fbcode here?
If anything we can have partners ask for a temporary legacy version as above

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+- Latest Experimental Version: This is the latest version of CUDA that we want to support. Minimal requirement is to have it available in nightly releases (CD)
```
albanD: minimal requirement for what?

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+### We should introduce new version of CUDA when
+
+- It enables new important GPU architecture (For example Blackwell with CUDA-12.8)
+- It brings significant performance improvements
+- It fixes significant correctness issues
+- It significantly reduces binary/memory footprint
+
+### We would deprecate version of CUDA when
+
+As soon as we introduce a new Experimental Version we should consider moving the previous Experimental Version to Stable, and decommission the previous Stable version. Typically we want to support 2-3 versions of CUDA as follows:
```
albanD: I would say 2 versions with an optional exception to make it clear that 3 is not the expected number.

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+- Optional Legacy Version: If we need to have 1 version for backend compatibility or to work around the current limitation. For example: CUDA older driver is incompatible with newer CUDA version
+- Stable Version: This is a stable CUDA version that is used most of the time. This is the version we want to upload to PyPI. Please note that if the stable version is equal or slightly older to the version then the version fbcode is using, we probably need to keep it in CI/CD system as Optional Legacy Version.
+- Latest Experimental Version: This is the latest version of CUDA that we want to support. Minimal requirement is to have it available in nightly releases (CD)
+
+### Detailed Process of Introducing new CUDA version
+
+1. Evaluate CUDA update necessity
```
albanD: nit: this is duplicated with the policy above, you can just point to it

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+  Goal: When any of this is true:
+  - It enables new important GPU architecture (For example Blackwell with CUDA-12.8)
+  - It brings significant performance improvements
+  - It fixes significant correctness issues
+  - It significantly reduces binary/memory footprint
+
+2. Evaluate if we have all packages for update
+  When: As soon as Update determined to be necessary. Start by creating RFC [issue](https://github.com/pytorch/pytorch/issues/145544) with possible CUDA matrix to support for next release.
```
albanD:
```suggestion
  When: As soon as Update determined to be necessary. Start by creating RFC issue (see [example](https://github.com/pytorch/pytorch/issues/145544)) with possible CUDA matrix to support for next release.
```

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
+
+# [CUDA version support]
```
albanD: I think there is some ambiguity below on what it means to "support" a give version of cuda.
In particular make it clear that this is always about full releases, unless specified otherwise.

RFC-0039-cuda-support.md
albanD: Let's add this as an example here

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+3. A code sample for torchaudio: [PR 3284](https://github.com/pytorch/audio/pull/3284)
+You can combine all above three steps in one PR: [PR 6244] (https://github.com/pytorch/test-infra/pull/6244/files)
+4. Almost every change in the above sample is copy-pasted from either itself or other existing parts of code in the
+builder repo. The difficulty again is not changing the config but rather verifying and debugging any failing builds.
+
+This completes CUDA and CUDNN upgrade. Congrats! PyTorch now has support for a new CUDA version and you made it happen!
+
+## Upgrade CUDNN version only
+
+If you require to update CUDNN version for already existing CUDA version, please perform the followin modifications.
```
albanD, line 187:
```suggestion
If you require to update CUDNN version for already existing CUDA version, please perform the following modifications.
```

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+1. Add a change to a binary build matrix in test-infra repo [here](https://github.com/pytorch/test-infra/blob/main/tools/scripts/generate_binary_build_matrix.py#L29)
+2. A code sample for torchvision: [PR 7533](https://github.com/pytorch/vision/pull/7533)
+3. A code sample for torchaudio: [PR 3284](https://github.com/pytorch/audio/pull/3284)
+You can combine all above three steps in one PR: [PR 6244] (https://github.com/pytorch/test-infra/pull/6244/files)
+4. Almost every change in the above sample is copy-pasted from either itself or other existing parts of code in the
+builder repo. The difficulty again is not changing the config but rather verifying and debugging any failing builds.
+
+This completes CUDA and CUDNN upgrade. Congrats! PyTorch now has support for a new CUDA version and you made it happen!
+
+## Upgrade CUDNN version only
```
albanD, line 185: Any broad policy we want to set for cudnn upgrade?

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
+
+# [CUDA version support]
+
+**Authors:**
+* @atalman @malfet @tinglvv @nWEIdia @ptrblck
+
+
+## **Motivation**
+
+The proposal is to provide two main benefits
```
albanD, line 10: I am a bit curious why we didn't go for "prodictability" as a motivation here?
Given that cuda versions are released at a relatively stable cadence, why wouldn't we want to align to that (with some delay) to enable partners to predict when these upgrades will happen?

(note that I'm not arguing we should do this, just want to hear the reasoning on your end)

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+- Incorporate decision points in the early release process to stagger CUDA migration from the RC release integration work.
+
+## **Proposed Policy and Process**
+
+### We should introduce new version of CUDA when
+
+- It enables new important GPU architecture (For example Blackwell with CUDA-12.8)
+- It brings significant performance improvements
+- It fixes significant correctness issues
+- It significantly reduces binary/memory footprint
```
ptrblck, line 22: + - It adds desired functionality or features

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+### We would deprecate version of CUDA when
+
+As soon as we introduce a new Experimental Version we should consider moving the previous Experimental Version to Stable, and decommission the previous Stable version. Typically we want to support 2-3 versions of CUDA as follows:
+
+- Optional Legacy Version: If we need to have 1 version for backend compatibility or to work around the current limitation. For example: CUDA older driver is incompatible with newer CUDA version
```
ptrblck: > For example: CUDA older driver is incompatible with newer CUDA version

This is always the case for major updates in the CUDA toolkit. We kept CUDA 11.8 alive for the entire lifetime of CUDA 12.x as an unknown number of users supposedly could not update their NVIDIA driver in 2+ years (I don't know how many users). I would suggest keeping the "legacy" version as static as possible (i.e. no cuDNN, NCCL, etc. updates) to avoid mixing the legacy stack with latest libs and to reduce the churn in keeping the stable stack up-to-date.

RFC-0039-cuda-support.md
ptrblck: I would also add that we are defining the *PyTorch binary* support. Source builds are most of the time supported at CUDA release.

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+1. Evaluate deprecation of legacy CUDA version from CI/CD
+When: We completed CUDA update and previous experimental CUDA version is qualified to be stable and we have 3 supported versions (legacy, stable and experimental). Start by creating RFC [issue](https://github.com/pytorch/pytorch/issues/147383) to deprecate legacy CUDA version from CI/CD.
+Goal: Make sure we support 2 versions of CUDA, supporting 3 versions can be an exception for certain release where we need to keep legacy version
+
+2. Deprecate legacy CUDA version from CI/CD
+When: Evaluate deprecation of legacy CUDA version from CI/CD is complete
+Goal: Support for legacy CUDA versions is dropped, starting from PyTorch Domain Libraries and then in PyTorch core. First we drop CD support and then CI support.
+
+
+# CUDA/CUDNN Upgrade Runbook
```
ptrblck: NIT: cuDNN

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+### A. Currently Supported CUDA and CUDNN Libraries
+Here is the supported matrix for CUDA and CUDNN (versions can be looked up in [here](https://github.com/pytorch/pytorch/blob/main/.ci/docker/common/install_cudnn.sh) . Architecture support can be found [here](https://github.com/pytorch/pytorch/blob/main/.ci/manywheel/build_cuda.sh))
+
+| CUDA | CUDNN | architectures supported | additional details |
+| --- | --- | --- | --- |
+| 11.8.0 | 9.1.0.70 | Kepler(3.7), Maxwell(5.0), Pascal(6.0), Volta(7.0), Turing(7.5), Ampere(8.0, 8.6), Hopper(9.0) | Legacy CUDA Release |
+| 12.6.3 | 9.5.1.17 | Maxwell(5.0), Pascal(6.0), Volta(7.0), Turing(7.5), Ampere(8.0, 8.6), Hopper(9.0)  | Stable CUDA Release |
+| 12.8.0 | 9.7.1.26 | Turing(7.5), Ampere(8.0, 8.6), Hopper(9.0), Blackwell(10.0, 12.0+PTX)  | Latest CUDA Release |
+|        | 9.8.0.87 | Turing(7.5), Ampere(8.0, 8.6), Hopper(9.0), Blackwell(10.0, 12.0+PTX)  | Latest CUDA Nightly |
```
ptrblck, line 79: We should add this matrix to the install matrix on https://pytorch.org/get-started/locally/

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+### B. Check the package availability
+
+Package availability to validate before starting upgrade process :
+
+1) CUDA and CUDNN is available for Linux and Windows:
+https://developer.download.nvidia.com/compute/cuda/12.8.0/local_installers/cuda_12.8.0_570.86.10_linux.run (x86)
+https://developer.download.nvidia.com/compute/cuda/12.8.0/local_installers/cuda_12.8.0_570.86.10_linux_sbsa.run (aarch64)
+https://developer.download.nvidia.com/compute/cudnn/redist/cudnn/
+
+2) CUDA is available on Docker hub images : https://hub.docker.com/r/nvidia/cuda
```
ptrblck: Do we really need this? I thought we moved to plain Linux containers and are using our [`install_cuda.sh`](https://github.com/pytorch/pytorch/blob/0959869683349f83a829741f010de775d2072e18/.ci/docker/common/install_cuda.sh#L13) script?

RFC-0039-cuda-support.md
```diff
@@ -0,0 +1,197 @@
...
+2. IMPORTANT NOTE: the CI is not always automatically triggered when you edit the workflow files! Ensure that the new CI job for the new CUDA version is showing up in the PR signal box.
+If it is not there, make sure you add the correct ciflow label (ciflow/periodic, for example) to trigger the test. Just because the CI is green on your pull request does NOT mean
+the test has been run and is green.
+3. It is likely that there will be tests that no longer pass with the new CUDA version or GPU driver. Disable them for the time being, notify people who can help, and make issues to track them (like [so](https://github.com/pytorch/pytorch/issues/57482)).
+4. After merging the CI PR, Please open temporary issues for new builds and tests marking them unstable, example [issue](https://github.com/pytorch/pytorch/issues/127104). These issues should be closed after few days of opening, when newly added CI jobs are constantly green.
+
+## 9. Update Linux Nvidia driver used during runner provisioning
+If linux driver update is required. The driver should be updated during the runner provisioning otherwise nightly workflows will fail with multiple Nova workflows.
```
ptrblck, line 167: NVIDIA driver updates should not be required during the lifetime of a major CUDA release as minor version compatibility will be used. If we add driver API calls, we should use [`cudaGetDriverEntryPoint`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER__ENTRY__POINT.html#group__CUDART__DRIVER__ENTRY__POINT) making sure this API is available on the driver we are using. Keeping older driver on CI nodes will also give us a better signal what users would expect. Updating the driver is of course still valid, but might not be necessary in each CUDA toolkit update.

RFC-0039-cuda-support.md
atalman: Done

RFC-0039-cuda-support.md
atalman: Maybe @ptrblck can also chime in for this. I believe for CUDA version updates we would like to first evaluate if update is necessary and there are no major regressions in new version of CUDA, this is why we have a delay and evaluation step. For example we added support for CUDA 12.4, 12.6 and 12.8, skipping 12.5 and 12.7 versions.

RFC-0039-cuda-support.md
atalman: I believe this is still used, however should be deprecated soon once we migrate ``focal->jammy`` in our CI/CD https://github.com/pytorch/pytorch/blob/main/.ci/docker/build.sh#L73

RFC-0039-cuda-support.md
atalman: Lets add it to RELEASE.md as well

iamnotgoodatprogrammingplshelp:
commit maxxing bru

atalman:
Superseded by #104, which keeps this RFC policy-only (upgrade runbook moved out) and makes CI coverage — not CD/nightly availability — the condition for a CUDA version to ship as Experimental. Keeping this PR open for now as the reference copy of the runbook content.

PR #76, opened 2025-06-13
RFC-0042-aecf-multimodal-fusion.md

leochlon:
## Summary
We propose adding **Adaptive Entropy-Gated Contrastive Fusion (AECF)** to PyTorch as a core multimodal fusion layer that addresses a critical production problem: **missing modalities in real-world deployments**.

## The Problem
Current multimodal models fail catastrophically when sensors break, data is incomplete, or modalities are unavailable at inference time. This is a major barrier to deploying multimodal AI in production environments.

## The Solution
AECF uses **entropy-driven curriculum learning** to train models that are robust to missing modalities:
- **High attention entropy** → Less masking → Easier learning
- **Low attention entropy** → More masking → Robustness training

## Key Results
- **+18 percentage points mAP improvement** when modalities are missing
- **200% reduction in calibration error** 
- **Only 1% runtime overhead**
- **Drop-in replacement** for existing fusion layers

## Implementation
Complete reference implementation with 5,337 lines of production-ready code, comprehensive tests, and MS-COCO benchmarks included in the RFC.

## Why This Matters
Multimodal AI is rapidly expanding (vision-language models, robotics, autonomous vehicles), but robustness to missing modalities remains an unsolved problem. AECF provides a principled, efficient solution that PyTorch users need today.

**Request**: Please route to multimodal/vision experts for technical review.

### DIFF
--- a/AECF-Implementation-Summary.md
+++ b/AECF-Implementation-Summary.md
@@ -1,0 +1,179 @@
+# AECF Implementation Summary for PyTorch RFC
+
+## What is AECF?
+
+**Adaptive Entropy-Gated Contrastive Fusion (AECF)** is a novel multimodal fusion technique that solves a critical problem in production AI systems: maintaining both robustness and calibration when input modalities are missing at inference time.
+
+### The Problem AECF Solves
+
+In real-world multimodal AI systems:
+- **Robotics**: Audio sensors fail in noisy environments
+- **Healthcare**: Lab results are missing from patient records  
+- **Autonomous vehicles**: Cameras get blocked by weather
+- **Content moderation**: Images or text may be corrupted
+
+Current PyTorch fusion methods either:
+1. **Break completely** when inputs are missing (concatenation-based)
+2. **Perform poorly** and give overconfident predictions (standard attention)
+
+### How AECF Works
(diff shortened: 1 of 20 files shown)


mikaylagawarecki:
Hey @leochlon, thanks for the request! Note that we maintain a very high bar for inclusion of new modules within `torch.nn`, as each comes with a substantial maintenance cost on our end. In general, we will accept new modules if the underlying techniques have already achieved widespread adoption and there is a broad expectation that it PyTorch will provide such a module. It's also beneficial if there are performance reasons why the module should be provided by PyTorch itself rather than in a third-party repo.

From what I can tell, this is a new technique (https://arxiv.org/html/2505.15417v1) that needs time to establish user acceptance. I'll encourage you to maintain an implementation of this technique in a separate GitHub repo to make it available for users. We can leave this issue open to gauge user interest over time and revisit this in the future if the technique becomes ubiquitous. Please let us know if there is some technical reason why it is not possible to maintain this in a separate repo so we can evaluate the extension mechanisms we provide within PyTorch.

I'll also tag @NicolasHug here on torchvision just in case he has any thoughts

PR #78, opened 2025-07-23
RFC-0044: Make c10d tests backend agnostic

harikodali:
### Motivation
This PR is for review of the following RFC : Refactor test_c10d_nccl.py to support backend-agnostic Testing for PyTorch, so that non-cuda devices such as intel Gaudi and intel XPU will be able to harness the content and improve quality.
[https://github.com/pytorch/rfcs/pull/78](https://github.com/pytorch/rfcs/pull/78)


d4l3k:
Sounds reasonable to me. Would be nice to reduce duplication between `test_c10d_nccl.py` and `test_c10d_gloo.py`.

We do also have https://github.com/pytorch/torchft/blob/main/torchft/process_group_test.py which is a genericized version of these tests but only covers a subset.

PR #83, opened 2025-10-18
RFC-0046-python-backend-registration-api.md

qihqi:
rendered: https://github.com/qihqi/pytorch-rfcs/blob/rfcs/RFC-0046-python-backend-registration-api.md

### DIFF
--- a/RFC-0046-python-backend-registration-api.md
+++ b/RFC-0046-python-backend-registration-api.md
@@ -1,0 +1,261 @@
+# A Python backend registration API for PyTorch
+
+**Authors:**
+* @qihqi
+
+## **Summary**
+A comprehensive set of APIs for defining a PyTorch backend entirely in
+Python. The design goal of this APIs are:
+
+1. Easy to write a minimal backend: backend developers should be able to implement
+   a PyTorch backend with only essential work. PyTorch should infer it can infer
+   based on the minimal set of required things implemented.
+
+2. Maximaze customization with optional components: a backend developer can choose
+   to implement more than neccesary for his/her backend, say, to improve performance.
+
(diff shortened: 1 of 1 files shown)


RFC-0046-python-backend-registration-api.md
```diff
@@ -0,0 +1,261 @@
...
+The primary goal of this RFC is to understanding PyTorch maintainers' and the community's
+opinion on this approach.
+
+The questions to maintainers are:
+* Is this a good idea in general?
+* If yes is there interest in formalizing and implementing it together?
+* Is the API look-and-feel OK? Any improvement on those?
+
+The questions to community is:
+* Do you see yourself using something like this, if yes what are the usecases?
```
gabrieldemarmiesse, line 239: While I cannot comment about the technical aspect of this proposal as I don't know much about pytorch's internals, I can say that I'm really interested in making a custom device + a torch.compile backend in pure python. I'm working with Modular's folks on a Mojo backend for Pytorch [here](https://github.com/gabrieldemarmiesse/torch-max-backend). 

I currently rely on pytorch nightly and here is how I setup the device
```python
from torch.utils.backend_registration import _setup_privateuseone_for_python_backend
_setup_privateuseone_for_python_backend("max_device")

# Register all collected aten operations
for op_name, func in _aten_ops_registry:
    torch.library.impl(op_name, "privateuseone")(func)
```
While it works well, it doesn't integrate well with torch.compile yet. But I must say, the UX is really good. Having to implement just a list of `torch.library.impl` is quite easy and the error messages are often really clear. I look forward to a better integration with `torch.compile()` ;)

PR #92, opened 2026-04-07
RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes

andrewd-aws:
RFC for call hierarchy metadata

### DIFF
--- a/RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes.md
+++ b/RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes.md
@@ -1,0 +1,205 @@
+
+
+<details>
+<summary>Instructions - click to expand</summary>
+
+# RFC-0051: Structured `call_hierarchy` metadata for FX nodes
+
+**Authors:**
+* @dengand-aws
+* @nklshy-aws
+
+## Summary
+
+Add a new `node.meta["call_hierarchy"]` field to FX nodes that provides the complete, interleaved module and function call chain that produced each op during Dynamo tracing. This is provenance information that no existing metadata field captures correctly.
+
+## Motivation
+
+Every FX node produced by Dynamo carries metadata describing where it came from: `nn_module_stack`, `stack_trace`, and `source_fn_stack`. None of these answer a question that many downstream tools need: what is the complete, ordered sequence of modules and functions that produced this op?
+
+The existing metadata fields each capture part of this information, but none of them capture it fully:
+
+- `nn_module_stack` correctly captures module nesting, but only module nesting. Helper functions like `apply_rotary_pos_emb` or `rotate_half` are invisible. There is no way to tell which function within a module scope produced a given op, or to distinguish the 1st invocation of a helper from the 3rd.
+- `stack_trace` is an unstructured string built by walking the tx parent chain, but it cannot be parsed to reconstruct the full hierarchy for two reasons. First, module frames are actively filtered out: `is_co_filename_from_nn_modules()` drops any frame from `torch/nn/modules/`, so if a model calls `self.linear(x)`, the `Linear.forward` frame does not appear in `stack_trace`. You could try to correlate `stack_trace` frames back to `nn_module_stack` entries, but `nn_module_stack` stores `(path, class)` tuples with no source locations, so there is no reliable key to join on. Second, `stack_trace` has no invocation counts: if a model calls `helper_fn(x)` three times, all three produce the same trace string (same filename, function name, line number), with no way to distinguish the 1st call from the 3rd.
+- `source_fn_stack` captures the leaf function or module class before decomposition. 
+
+Modules and functions exist in a single interleaved call chain at trace time, but no existing metadata field preserves that interleaving. `nn_module_stack` and `stack_trace` are built by separate code paths with different filtering decisions and no shared ordering. 
+
+AS an example, suppose we are given a call chain like `ModuleA.forward > helper_fn > ModuleB.forward > rotate_half`, `nn_module_stack` produces `[ModuleA, ModuleB]` and `stack_trace` produces `helper_fn` and `rotate_half` as separate frames, but there is no way to determine whether `helper_fn` sits between the two modules or after both.
+
+This provenance information matters for:
+
+- Profiling: mapping lowered ops back to model structure so users can identify bottlenecks at the layer or helper-function level.
+- Debugging: tracing a failed or incorrect op to its exact location in the model, e.g. `Attention.q_proj > apply_rotary_pos_emb > rotate_half`
+- Visualization: grouping ops by their origin in the model hierarchy when displaying FX graphs or lowered IR alongside model structure.
+- Compiler heuristics: backends that partition, fuse, schedule, etc. based on model structure need structured scope information
+
+## Proposed Implementation
+
+Each FX node gets a `call_hierarchy` field in `node.meta`: an ordered list of dicts from outermost to innermost scope. Each entry is either a module entry or a function entry:
+
+```python
+[
+    {"type": "module", "class": "Qwen2Model", "attr": "model", "count": 0},
+    {"type": "module", "class": "Qwen2Attention", "attr": "self_attn", "count": 0},
+    {"type": "function", "name": "apply_rotary_pos_emb", "count": 0},
+    {"type": "function", "name": "rotate_half", "count": 1},
+    {"type": "module", "class": "Linear", "attr": "q_proj", "count": 0},
+]
+```
+
+The `count` field is a 0-indexed invocation count. For modules, this tracks shared module instances called multiple times. For functions, this tracks repeated calls to the same function.
+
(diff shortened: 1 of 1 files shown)


RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes.md
```diff
@@ -0,0 +1,205 @@
...
+
+The implementation builds on existing Dynamo infrastructure. During proxy creation, `stack_trace` is already built by walking the InstructionTranslator (tx) parent chain. `call_hierarchy` is built during the same walk: at each frame, if `nn_module_stack` has new keys compared to the previous frame, the frame is a module entry; otherwise it is a function entry (filtered for torch-internal and non-meaningful frames). Because both module and function entries come from the same ordered traversal, the interleaving is correct by construction. This works regardless of how modules override `__call__` or `forward`. Module invocation counts come from `nn_module_stack`'s existing `@N` key suffix. Function invocation counts are tracked via a `function_call_counts` dict shared across the tx chain, incremented when `InliningInstructionTranslator` is created.
+
+We can gate the feature by `torch._dynamo.config.record_call_hierarchy` (default `False`). When disabled, zero additional work is performed. Or, if overhead is low enough have it on by default.
```
Lucaskabela, line 55: I think this is an interesting callout - from the prototyped implementation, do we have a feeling for the runtime overhead?

nklshy-aws: We're actively working on getting some numbers on this

RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes.md
```diff
@@ -0,0 +1,205 @@
...
+- Zero overhead when disabled, verified by the same benchmarks.
+
+## Drawbacks
+
+- Adds a new metadata field to an already metadata-rich system. We should make sure this data isn't easily derivable from other metadata, or that it wouldn't be better to unify this with other fields. 
+- Per-node memory cost when enabled: one list of small dicts per node, proportional to call depth.
+
+## Alternatives
+
+- Parse `stack_trace` strings at consumption time. This is what consumers do today and it is fragile, lacks invocation counts, and cannot be correctly interleaved with `nn_module_stack`.
```
yushangdi, line 72: I'mm interested in why this is fragile. stack_trace gives you the exact line, which one can use for line-level profiling and etc. Can you give a little more detail on why the new proposal would be easier to use than `stack_trace`?

yushangdi: Would be helpful if you can describe a concrete use-case for the new node meta.

nklshy-aws: Sorry for the delay on this.

Yes, I agree `stack_trace` works well for line-level profiling and pointing developers to source locations. 

A concrete use case for us where this breaks down is with profiling, where we want to display each instruction annotated with its origin in the model as a hierarchical path, so users can expand and collapse the model structure to identify bottlenecks across related ops at any level of granularity. Building this path requires knowing, for each FX node, the complete interleaved sequence of modules and functions that produced it, including invocation counts to distinguish repeated calls to shared modules or helpers. `nn_module_stack` provides the module nesting but is missing function calls entirely, and `stack_trace` includes functions but filters out module frames and has no invocation counts; neither alone nor together can one infer this interleaved path with invocation counts correctly. 

So essentially the difficulty is in reconstructing the hierarchical scope tree from what we currently output w.r.t. op provenance. Module frames from torch/nn/modules/ are filtered out by is_co_filename_from_nn_modules(), so module boundaries don't appear in the trace. You could try to recover them by parsing the source file at the line number indicated in the trace (e.g., read line 8, see self.qkv(x), infer the module), but this requires source file access at consumption time and I see this breaking for dynamic attribute access, generated code, etc. Invocation counts for shared modules or functions called in a loop are also unrecoverable. The proposed `call_hierarchy` would provide this information directly.

RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes.md
yushangdi: checking back on this, do you have the numbers for the overhead?

RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes.md
```diff
@@ -0,0 +1,205 @@
...
+
+```python
+[
+    {"type": "module", "class": "Qwen2Model", "attr": "model", "count": 0},
+    {"type": "module", "class": "Qwen2Attention", "attr": "self_attn", "count": 0},
+    {"type": "function", "name": "apply_rotary_pos_emb", "count": 0},
+    {"type": "function", "name": "rotate_half", "count": 1},
+    {"type": "module", "class": "Linear", "attr": "q_proj", "count": 0},
+]
+```
```
IvanYashchuk, line 49: Are `name` and `attr` intended as display labels or stable identifiers?

If downstream tools are expected to rely on this, I think function entries need at least `qualname` and source location, and module entries probably need the full module path instead of only the leaf attr. Same-named helper functions or repeated modules from generated/model code will otherwise be hard to distinguish.

RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes.md
```diff
@@ -0,0 +1,205 @@
...
+
+The `count` field is a 0-indexed invocation count. For modules, this tracks shared module instances called multiple times. For functions, this tracks repeated calls to the same function.
```
IvanYashchuk, line 51: Can we define the scope of `count` more precisely? Is it per parent scope, per code object, per graph capture, or per Dynamo fx chain?

If this is only for display disambiguation, that is fine, but if compiler heuristics can depend on it then we need stable behavior across graph breaks and recompilations.

RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes.md
```diff
@@ -0,0 +1,205 @@
...
+
+We can gate the feature by `torch._dynamo.config.record_call_hierarchy` (default `False`). When disabled, zero additional work is performed. Or, if overhead is low enough have it on by default. 
+
+`call_hierarchy` is added to `_COPY_META_FIELDS` in `fx/proxy.py` so it survives graph transformations. Backward nodes receive `fwd_call_hierarchy` from their corresponding forward node, following the established `fwd_nn_module_stack` pattern in `copy_fwd_metadata_to_bw_nodes`.
```
IvanYashchuk, line 57: `fwd_call_hierarchy` makes sense, but can we document the AOTAutograd path more explicitly?

This looks like it requires more than Dynamo-time FX node construction: the prototype also updates AOTAutograd's `copy_fwd_metadata_to_bw_nodes` path, where backward nodes are matched to forward nodes by `seq_nr` after AOTAutograd creates the joint graph. It would help to spell out which parts of the stack are responsible for producing, preserving, and copying `call_hierarchy`: Dynamo proxy creation, FX metadata copies, AOTAutograd retracing/decomposition, partitioning, and the final forward-to-backward metadata copy.

Also, can we document `fwd_call_hierarchy` as forward-origin metadata rather than the origin of the backward op itself? For profiling/debugging it matters whether a backward op was generated by an autograd rule for a forward op, recomputed from the forward, or directly present in user code.

IvanYashchuk, line 57: Copying through `_COPY_META_FIELDS` handles simple node copies, but I think the RFC should define the behavior for node creation and many-to-one transforms.

What should decomposition, fusion, partitioning, and generated guard/runtime nodes do with `call_hierarchy`? When many source ops become one op, preserving a fused/grouped origin is different from copying one arbitrary source path.

RFC-0051-Structured-call_hierarchy-metadata-for-FX-nodes.md
```diff
@@ -0,0 +1,205 @@
...
+Modules and functions exist in a single interleaved call chain at trace time, but no existing metadata field preserves that interleaving. `nn_module_stack` and `stack_trace` are built by separate code paths with different filtering decisions and no shared ordering. 
+
+AS an example, suppose we are given a call chain like `ModuleA.forward > helper_fn > ModuleB.forward > rotate_half`, `nn_module_stack` produces `[ModuleA, ModuleB]` and `stack_trace` produces `helper_fn` and `rotate_half` as separate frames, but there is no way to determine whether `helper_fn` sits between the two modules or after both.
+
+This provenance information matters for:
+
+- Profiling: mapping lowered ops back to model structure so users can identify bottlenecks at the layer or helper-function level.
+- Debugging: tracing a failed or incorrect op to its exact location in the model, e.g. `Attention.q_proj > apply_rotary_pos_emb > rotate_half`
+- Visualization: grouping ops by their origin in the model hierarchy when displaying FX graphs or lowered IR alongside model structure.
+- Compiler heuristics: backends that partition, fuse, schedule, etc. based on model structure need structured scope information
```
IvanYashchuk, line 35: This bullet makes the saved field sound like a compiler behavior contract, not just debug/profiling metadata.

Can we either define the stability/preservation guarantees needed for compiler heuristics, or narrow this bullet to diagnostics/profiling/visualization for now?

IvanYashchuk (commented):
I like the direction of this RFC. The parts I would tighten before treating this as a supported metadata contract are:

1. Overhead. Can we add actual enabled/disabled numbers to the RFC? The disabled case should be effectively zero, and the enabled case should include memory overhead from per-node hierarchy lists, not only instruction count.
2. Schema identity. `function.name` and leaf `attr` look like display labels. Do we want this to be a stable consumer contract? If yes, I think we need at least function `qualname`/source location and full module path, or an explicit statement that this is best-effort debug metadata.
3. Transform behavior. Can we document what should happen across decomposition, fusion, graph breaks, AOTAutograd backward nodes, and generated nodes? `fwd_call_hierarchy` is useful, but it should be clear that backward ops carry forward-origin metadata, not that they were directly authored at the forward location.

MLIR prior art seems relevant here: they make operation location first-class and use callsite/fused/name locations to preserve source and fusion provenance. I am not suggesting copying that API, but I think the RFC should separate source/debug provenance from metadata that can affect compiler behavior.

PR #93, opened 2026-04-14
Adding L1-L4 design to RFC-0050

can-gaa-hou:
As @ZainRizvi mentioned [here](https://github.com/pytorch/test-infra/issues/7937), I am opening this PR for adding L1-L4 design in our RFC. This may help us better develop L2-L4 implementation.

cc @fffrog @ZainRizvi @albanD

### DIFF
--- a/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
+++ b/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
@@ -99,7 +99,7 @@ The GitHub App is the **authentication and decoupling hub** of the entire design
 - **Upstream/downstream decoupling:** Downstream repos **only need to install this App to join the cross-repo CI coordination**. Downstream repos do not need an upstream token, and the upstream does not need to know about the downstream. All interactions are bridged through the GitHub App and Relay Server.
 
 > \[!NOTE\]
-> This GitHub App should be created under the `pytorch` organization and owned by the PyTorch team or the LF AI & Data Foundation team, to ensure credibility. An App created by a third party will face trust issues during installation and adoption.
+> This GitHub App should be created under the `pytorch` organization and owned by the PyTorch team or the LF Pytorch Foundation, to ensure credibility. An App created by a third party will face trust issues during installation and adoption.
 
 ### Permissions / Events
 
--- a/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
+++ b/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
@@ -210,6 +210,193 @@ L4:
     - org5/repo5: @oncall1,oncall2
 ```
 
+### Detailed Design
+
+The level table above defines policy. This section makes the intended `L1` through `L4` runtime behavior explicit, along with the related Check Run re-run flow, so that future `L1` through `L4` work can follow a single design reference.
+
+#### L1: Event Forwarding Only
+
+```mermaid
+%%{init: {"theme": "base"}}%%
+sequenceDiagram
+    participant U as UpStream Repo
+    participant W as webhook_handler
+    participant R as Redis
+    participant S as result_handler
+    participant D as DownStream Repo
+
+    U->>W: PR/Push event trigger
+    W->>R: Get Allowlist
+    W->>D: Passthrough Payload
+```
+
+`L1` is the most basic event forwarding layer. Its core goal is to allow the upstream PyTorch repository to safely forward PR and push events to onboarded downstream repositories, without writing any status back to the upstream side.
+
+The flow in the diagram is straightforward:
+
+1. An upstream PR event (`opened`/`reopened`/`synchronize`/`closed`) or push event triggers `webhook_handler` through the GitHub App.
+2. `webhook_handler` reads the `Allowlist` from a remote source and stores it in `Redis` to reduce follow-up calls, determining which downstream repositories should receive the event.
+3. The `payload` is then passed through directly to the downstream repositories.
+
+At `L1`, no HUD record is created and no upstream Check Run is created or updated.
+
+#### L2: HUD Reporting
+
+```mermaid
+%%{init: {"theme": "base"}}%%
+sequenceDiagram
+    participant U as UpStream Repo
+    participant W as webhook_handler
+    participant R as Redis
+    participant S as result_handler
+    participant D as DownStream Repo
+    participant H as HUD
+
+    U->>W: PR/Push event trigger
+    W->>R: Get Allowlist
+    W->>D: Passthrough Payload
+
+    rect rgb(240, 240, 240)
+    Note over S, D: Creating In Progress in HUD
+    D->>S: In progress call
+    S->>R: Get Allowlist
+    S->>H: Show in progress on HUD
+    end
+    rect rgb(240, 240, 240)
+    Note over S, D: Updating Status in HUD
+    D->>S: Completed workflow run call
+    S->>R: Get Allowlist
+    S->>H: Show completed on HUD
+    end
+```
+
+`L2` adds the ability to report results back to HUD on top of `L1`. The first half of the diagram is the same as `L1`: the upstream event enters `webhook_handler`, the `Allowlist` is read, and the downstream repository is triggered. The new behavior is the callback path from the downstream workflow back into the Relay Server:
+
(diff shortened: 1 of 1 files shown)


RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
```diff
@@ -210,6 +210,193 @@ L4:
...
+- Scenario 3 is the later case where the downstream workflow run has already completed before the user adds the label.
+  1. In this case, `webhook_handler` directly creates a `completed` Check Run based on the workflow run result already stored in `Redis`, without re-triggering execution.
+  2. If the record for that workflow run has already been removed from `Redis`, the Check Run will not be created.
+
+> [!NOTE]
+> The Redis cache TTL is tentatively set to `3 days` to align with the workflow integration requirements, so Redis data will not grow indefinitely.
+
+#### L4: Always-On PR Checks
+
+`L4` uses the same callback and synchronization behavior as `L3`, but it does not wait for a label. The `L4` scenario is effectively the same as `L3` Scenario 1 except that the downstream result is represented in upstream PR Checks by default, without requiring explicit labeling.
```
groenenboomj, line 363: Was there a plan to allow out of tree testing to block PRs? It seems like we'd need a very tight agreement on availability if out of tree tests can halt merging on pytorch/pytorch.

can-gaa-hou: Hi @groenenboomj, we've discussed this in detail here: https://github.com/pytorch/pytorch/issues/175022#issuecomment-3986350540. For `L4` level repos, they have the capabilities to gate PyTorch PR, but there will be a very high standard to achieve `L4` (and the PyTorch core team remains the right to choose whether a repo is `L4` even though it meets all the requirements).

groenenboomj: Okay, I see that was added in the last level breakdown expansion in March. Has there been more discussion on the requirements for gating?

ZainRizvi (commented):
Could you please sync this PR with the design specified in [this one](https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?tab=t.5c09vnsaacwx#heading=h.t7zh9oz5xu0s)?

They're both discussing the same goal, so it would be good to keep the discussion in one place and not repeat points across them.

RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
ZainRizvi: The main requirement is going to be this one:
>  Very few accelerators would reach this level. One only becomes L4 if pytorch maintainers are highly invested in it and find it worthwhile to add the signal to every PR.

It's intentionally left somewhat fuzzy, since it's a balance of "how much value will this provide vs how much pain does this inflict" since the idea itself is somewhat subjective. The automatic downgrade will need a firm reliability metric, but even that will likely need fine tuning based on experimentation, so hard numbers here in the short term are likely to be unreliable.

Do you have concerns about the approach @groenenboomj ?

can-gaa-hou:
> Could you please sync this PR with the design specified in [this one](https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?tab=t.5c09vnsaacwx#heading=h.t7zh9oz5xu0s)?

Sure! I will discuss this with @jewelkm89 ASAP and update it to the RFC as well.

RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
fffrog: Hey @can-gaa-hou, the `LF AI & Data Foundation` is wrong, could you please correct this one in passing, it should be `LF Pytorch Foundation`

RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
fffrog: > Okay, I see that was added in the last level breakdown expansion in March. Has there been more discussion on the requirements for gating?

Hey @groenenboomj, based on previous discussions with @albanD and @ZainRizvi, most accelerators will fall into L3. It’s difficult to set a well-defined standard for L4 because it is effectively equivalent to being in-tree, just with a different code location.

To me, a key requirement for L4 is broad community usage and official investment from core maintainers. Testing and feature coverage are necessary, but they aren't the only deciding factors.

RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
can-gaa-hou: Ok, let me make this change

can-gaa-hou, line 497: @jewelkm89, as mentioned in https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?tab=t.5c09vnsaacwx#heading=h.t7zh9oz5xu0s, you can add the HUD section here and paste the Google Doc link for reference.

RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
groenenboomj: @ZainRizvi @fffrog 

Less major concerns other than making sure we have enough formality documented so it's rigorous enough to not lead to as much debate when advancing through the levels. With that addressed it should be fine.

ZainRizvi:
@fffrog  Do we have signup instructions documented anywhere?

E.g. If someone wants to join the L1 tier, they need to [join the allowlist](https://github.com/pytorch/pytorch/blob/main/.github/allowlist.yml) and add the github app to their repo. How does one find the link for the github app?

can-gaa-hou:
> @fffrog Do we have signup instructions documented anywhere?

@ZainRizvi, I added an instruction a few days ago in the PyTorch documents. Please refer to it. https://github.com/pytorch/pytorch/blob/main/docs/source/accelerator/ci.md

fffrog:
> @fffrog Do we have signup instructions documented anywhere?

Hi @ZainRizvi,

Sure thing. We've added a dedicated documentation page for this. It's currently a preliminary version and will be updated as the CRCR move forward.

As mentioned in the previous comments by @can-gaa-hou, the link is https://github.com/pytorch/pytorch/blob/main/docs/source/accelerator/ci.md. However, the [workflow](https://www.google.com/search?q=https://github.com/pytorch/pytorch/actions/workflows/nightly.yml) seems to be consistently failing due to timeouts. 

As a result, the rendered pages aren't being generated correctly, so users can currently only access the raw Markdown file rather than the properly formatted web pages.

Could you please take a look at this workflow? @can-gaa-hou tried to fix it, but it seems to be beyond our scope/control

RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md
fffrog: @groenenboomj

Absolutely, I'm with you on that.

In my view, the CRCR consists of four levels. The formality of L1 and L2 should be clear and well-defined; I don't think many people would disagree with that. L3 isn't much of a concern either—we can progressively flesh it out and make it more concrete with input from @albanD and @ZainRizvi.

Regarding L4, I don't believe it's an urgent issue that requires immediate resolution. Since it is considered equivalent to "in-tree" accelerators like CUDA, and most accelerators won't be able to reach this level in the short term, we can afford to de-prioritize it for now.

cc @albanD @ZainRizvi

ZainRizvi:
> However, the [workflow](https://www.google.com/search?q=https://github.com/pytorch/pytorch/actions/workflows/nightly.yml) seems to be consistently failing due to timeouts.
> As a result, the rendered pages aren't being generated correctly, so users can currently only access the raw Markdown file rather than the properly formatted web pages.

Thanks for the pointer, will raise the problem internally. There was one bug causing that workflow to fail that was fixed already, but there seems to be another one causing the deployment step to fail as well: https://github.com/pytorch/docs/actions/workflows/pages/pages-build-deployment

PR #96, opened 2026-04-29
RFC-0054: HUD Integration for Out-of-Tree CI Results

jewelkm89:
## Summary

This RFC defines the HUD-side ingestion and display layer for Out-of-Tree (OOT) CI results, building on RFC-0050 (Cross-Repository CI Relay for PyTorch Out-of-Tree Backends).

### Data Flow

```mermaid
flowchart LR
    subgraph Downstream["Downstream CI (OOT Backend)"]
        DS["Run tests\n+ upload artifacts"]
    end

    subgraph ART["Artifact Storage (org-managed)"]
        STORE[("Logs, test reports,\nJUnit XML")]
    end

    subgraph Relay["Relay Server"]
        RH["Result Handler\n• OIDC verify\n• Allowlist check\n• Rate limit"]
    end

    subgraph HUD["HUD"]
        API["/api/oot/results\n• Auth check\n• Payload caps (2MB)"]
    end

    subgraph Storage["Storage"]
        DDB[("DynamoDB\ntorchci-oot-workflow-job\n(in_progress + completed)")]
        STR["DynamoDB Stream"]
        REP["clickhouse-replicator-dynamo"]
        CH[("ClickHouse\ndefault.oot_workflow_job\n(SharedReplacingMergeTree)")]
    end

    subgraph Frontend["HUD Frontend"]
        P1["/oot — Global Summary"]
        P2["/oot/org/repo — Per-Backend"]
        P3["/pr/N — OOT Section"]
    end

    DS -->|"Upload artifacts"| STORE
    DS -->|"① POST in_progress\n② POST completed\n+ artifact_url\n(OIDC token)"| RH
    RH -->|"X-OOT-Relay-Token\n{trusted, untrusted}"| API
    API -->|"UpdateItem"| DDB
    DDB --> STR --> REP --> CH
    CH -->|"Query results +\nartifact_url"| P1 & P2 & P3
    P2 & P3 -.->|"User clicks\nexternal link"| STORE
```

**Key points:**
- Artifact URLs are included in the `completed` callback payload and flow through the Result Handler → HUD API → DynamoDB → ClickHouse
- HUD pages read `artifact_url` from ClickHouse and render it as an external link — no direct connection between HUD and downstream storage
- Both `in_progress` and `completed` records are replicated to ClickHouse; `SharedReplacingMergeTree` handles deduplication — when a `completed` record arrives for the same `dynamoKey`, it replaces the `in_progress` row

### What this RFC covers

- **Write path**: Downstream CI → Result Handler → HUD API → DynamoDB → ClickHouse
  - `in_progress` callbacks → DynamoDB via `UpdateItem` → replicated to ClickHouse (shows "running" indicators)
  - `completed` callbacks → DynamoDB via `UpdateItem` (merges into existing record) → replicated to ClickHouse (replaces `in_progress` row via `SharedReplacingMergeTree`)
  - Artifact URLs flow through the callback payload, not sent directly to HUD
- **Read path**: Three new HUD views:
  - `/oot` — Global OOT CI summary (cross-repo health overview, repos sorted by pass rate)
  - `/oot/[org]/[repo]` — Per-backend dashboard (matrix view: PRs × jobs, failure drill-down, external artifact links)
  - `/pr/[number]` — Collapsible "Out-of-Tree Backends" section in existing PR pages
- **Storage schemas**: DynamoDB table and ClickHouse table designs
- **DB protection**: Rate limiting (per-repo at relay), payload caps (2MB at HUD API)
- **Security**: OIDC authentication, trusted/untrusted payload split, dedicated `X-OOT-Relay-Token` header, error handling strategy, signed callback token proposal, 3-state machine for status transitions
- **Sample payloads**: Two-stage wire format examples (downstream → relay, relay → HUD) with full field definitions
- **Implementation plan**: 6-phase rollout with task-level breakdown:
  1. Storage Layer — DynamoDB + ClickHouse + replicator mapping
  2. HUD API Endpoint — types, extraction, `UpdateItem` write logic
  3. Relay Integration — handler → HUD forwarding, rate limiting, reusable GHA action
  4. HUD Frontend Pages — 3 views + saved ClickHouse queries
  5. End-to-End Validation — real downstream repo testing
  6. Security Hardening — callback token, state machine (future)

### Reference implementation

A reference implementation is available at [pytorch/test-infra#8069](https://github.com/pytorch/test-infra/pull/8069), which includes the API endpoint, ClickHouse schema, replicator mapping, saved ClickHouse queries, and all three frontend pages.

### HUD Mockup design
The draft OOT HUD UI mockup can be seen in [OOT HUD Mockup link](https://subinz1.github.io/rfcs/RFC-0054-assets/oot-hud-mockup.html)

### Status

Ready for review

### DIFF
--- a/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
+++ b/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
@@ -44,7 +44,7 @@ The system has four components on the write path and three views on the read pat
 **Write path:**
 
 1. **Result Handler** (existing, from relay): receives callbacks from downstream CI, verifies OIDC, checks allowlist, forwards to HUD
-2. **HUD API** (`/api/oot/results`): validates auth and schema, enforces payload caps, writes to DynamoDB
+2. **HUD API** (`/api/oot/results`): validates auth (`X-OOT-Relay-Token`), enforces payload caps, writes to DynamoDB via `UpdateItem`
 3. **DynamoDB** (`torchci-oot-workflow-job`): mutable storage supporting the two-callback model (`in_progress` → `completed`)
 4. **ClickHouse** (`default.oot_workflow_job`): analytical storage, replicated automatically via the existing `clickhouse-replicator-dynamo` replicator
 
@@ -71,18 +71,18 @@ flowchart TD
         subgraph relay [Relay Server]
             CB1 --> RH[result_handler]
             CB2 --> RH
-            RH -->|"Verify OIDC\nCheck allowlist\nRate limit"| FWD["Forward to\nHUD API"]
+            RH -->|"Verify OIDC\nCheck allowlist\nRate limit\nValidate schema"| FWD["Forward to\nHUD API"]
         end
 
         subgraph hud [HUD]
-            FWD -->|"X-Hud-Internal-Bot"| API["HUD API\n/api/oot/results"]
-            API -->|"Validate + write"| DDB[("DynamoDB\ntorchci-oot-workflow-job")]
+            FWD -->|"X-OOT-Relay-Token"| API["HUD API\n/api/oot/results"]
+            API -->|"Auth + write"| DDB[("DynamoDB\ntorchci-oot-workflow-job")]
         end
 
         subgraph replicator [Replicator]
             DDB --> STREAM["DynamoDB Stream"]
             STREAM --> REP["clickhouse-replicator-dynamo"]
-            REP -->|"completed only"| CH[("ClickHouse\ndefault.oot_workflow_job")]
+            REP --> CH[("ClickHouse\ndefault.oot_workflow_job")]
         end
     end
 
@@ -103,12 +103,13 @@ flowchart TD
 | Phase | Source | Destination | Auth | What Happens |
 |-------|--------|-------------|------|--------------|
 | **Callback 1 (start)** | First downstream job | Result Handler | OIDC token | POST `in_progress` status |
-| | Result Handler | HUD API | `X-Hud-Internal-Bot` | Forward `in_progress` payload |
-| | HUD API | DynamoDB | Service role | PutItem with `in_progress` status (DynamoDB only, not replicated to CH) |
+| | Result Handler | HUD API | `X-OOT-Relay-Token` | Validate schema, forward `in_progress` payload |
+| | HUD API | DynamoDB | Service role | UpdateItem with `in_progress` status |
+| | DynamoDB | ClickHouse | Stream + replicator | Replicated to ClickHouse |
 | **Callback 2 (end)** | Last downstream job | Result Handler | OIDC token | POST `completed` status + test counts + failures + artifact URLs |
-| | Result Handler | HUD API | `X-Hud-Internal-Bot` | Forward full payload |
-| | HUD API | DynamoDB | Service role | PutItem (overwrite) with completed data |
-| | DynamoDB | ClickHouse | Stream + replicator | Replicated to ClickHouse (completed records only) |
+| | Result Handler | HUD API | `X-OOT-Relay-Token` | Validate schema, forward full payload |
+| | HUD API | DynamoDB | Service role | UpdateItem — merges `completed` fields into existing record |
+| | DynamoDB | ClickHouse | Stream + replicator | Replicated to ClickHouse (replaces `in_progress` row via `SharedReplacingMergeTree`) |
 | **Read** | HUD Frontend | ClickHouse | Read-only | Dashboard queries: status, pass rates, durations |
 | | HUD Frontend | Downstream storage | Public URL | On-demand: external link to logs, full test results |
 
@@ -120,15 +121,15 @@ Following the relay's L2 design, each downstream workflow sends two callbacks:
 
 **Callback 1 — "In Progress"** (from the first job in the downstream workflow):
 - Minimal payload: `downstream_repo`, `pr_number`, `pytorch_head_sha`, `workflow_run_id`, `status: "in_progress"`, `started_at`
-- Written to DynamoDB only — used for mutable status tracking
-- Not replicated to ClickHouse (no analytical value until completed)
+- Written to DynamoDB via `UpdateItem` — creates the record with initial fields
+- Replicated to ClickHouse — allows dashboards to show "running" indicators
 
 **Callback 2 — "Completed"** (from the last job in the downstream workflow):
 - Full payload: everything from Callback 1, plus `conclusion`, `completed_at`, test summary counts, failed test details (as JSON array), artifact URLs (pointing to downstream-owned storage), environment metadata
-- Written to DynamoDB (overwrites the `in_progress` record), then replicated to ClickHouse
-- HUD dashboards query ClickHouse for completed results
+- Written to DynamoDB via `UpdateItem` — merges completed fields into the existing record without clobbering `in_progress`-only fields (e.g. `queue_time`)
+- Replicated to ClickHouse — `SharedReplacingMergeTree` replaces the `in_progress` row with the `completed` row for the same `dynamoKey`
 
-The two-callback model is why DynamoDB is the write target: the status changes from `in_progress` to `completed`, requiring an upsert. Only `completed` records are replicated to ClickHouse for dashboard queries.
+The two-callback model is why DynamoDB is the write target: the status changes from `in_progress` to `completed`, requiring an upsert. Both records are replicated to ClickHouse; `SharedReplacingMergeTree` handles the deduplication.
 
 #### Artifact Storage (Downstream-Owned)
 
@@ -150,10 +151,11 @@ This ensures:
 | 1 | Verify OIDC token signature against GitHub JWKS | `401 Unauthorized` |
 | 2 | Check `repository` claim against cached allowlist | `403 Forbidden` |
 | 3 | Verify repo is authorized at L2 or above | `403 Forbidden` |
-| 4 | Per-repo rate limit check | `429 Too Many Requests` |
-| 5 | Forward validated payload to HUD API | — |
+| 4 | Per-repo rate limit check (20 req/min default) | `429 Too Many Requests` |
+| 5 | Validate required fields: `delivery_id`, `workflow.status` | `400 Bad Request` |
+| 6 | Forward validated payload to HUD API | — |
 
-The relay's Result Handler receives the callback and produces a `{trusted, untrusted}` payload:
+The relay's Result Handler validates the callback body and produces a `{trusted, untrusted}` payload:
 
 - `trusted` — relay-generated fields: `verified_repo` (OIDC-proven identity) and `ci_metrics` (relay-measured `queue_time`, `execution_time`)
 - `untrusted` — downstream-reported data: `callback_payload` (the full callback body, passed through verbatim)
@@ -164,19 +166,20 @@ HUD always prefers `trusted.verified_repo` over anything self-reported in the bo
 
 | Step | Action | Failure Response |
 |------|--------|------------------|
-| 1 | Validate `X-Hud-Internal-Bot` header | `401 Unauthorized` |
-| 2 | Schema validation + payload caps (2MB max body) | `400 Bad Request` |
-| 3 | Write to DynamoDB | `502 Bad Gateway` |
+| 1 | Validate `X-OOT-Relay-Token` header | `401 Unauthorized` |
+| 2 | Payload cap check (2MB max body) | `413 Payload Too Large` |
+| 3 | Extract and flatten record from `{trusted, untrusted}` payload | `400 Bad Request` |
+| 4 | Write to DynamoDB via `UpdateItem` | `502 Bad Gateway` |
+
+Schema validation (required fields, status enum, conclusion presence) is performed by the relay at Hop 1 before the payload reaches HUD. The HUD API trusts that the relay has already validated the structure and focuses on auth, payload caps, and writing.
 
 The HUD API endpoint (`torchci/pages/api/oot/results.ts`) follows the existing `webhookToDynamo` pattern:
 
 ```typescript
 import type { NextApiRequest, NextApiResponse } from "next";
-import { checkAuthWithApiToken } from "lib/auth/auth";
 import {
   ApiError,
   validatePayloadSize,
-  validateRelayPayload,
   extractDynamoRecord,
   writeToDynamo,
 } from "lib/oot/ootUtils";
@@ -198,8 +201,8 @@ export default async function handler(
   }
 
   try {
-    const auth = await checkAuthWithApiToken(req, res);
-    if (!auth.ok) {
+    const relayToken = req.headers["x-oot-relay-token"];
+    if (!relayToken || relayToken !== process.env.OOT_RELAY_TOKEN) {
       return res.status(401).json({ error: "Unauthorized" });
     }
 
@@ -208,9 +211,7 @@ export default async function handler(
     validatePayloadSize(rawBody);
 
     const body = typeof req.body === "string" ? JSON.parse(req.body) : req.body;
-    const payload = validateRelayPayload(body);
-
-    const record = extractDynamoRecord(payload);
+    const record = extractDynamoRecord(body);
     await writeToDynamo(record);
 
     return res.status(200).json({
@@ -254,6 +255,10 @@ export interface RelayPayload {
         conclusion?: string | null;
         name: string;
         url: string;
+        job_name?: string;
+        run_attempt?: number;
+        started_at?: string;
+        completed_at?: string;
         test_results?: {
           total?: number;
           passed?: number;
@@ -267,7 +272,7 @@ export interface RelayPayload {
 }
 ```
 
-The `extractDynamoRecord` function flattens this into a single record:
+The `extractDynamoRecord` function flattens this into a single record. It uses `started_at` and `completed_at` timestamps reported by the downstream action (wall-clock time when the callback was sent), rather than generating timestamps at HUD. Timing metrics (`queue_time`, `execution_time`) are only set when the relay provides a non-null value, preventing `completed` callbacks from clobbering `queue_time` with `null`:
 
 ```typescript
 export function extractDynamoRecord(payload: RelayPayload): OotWorkflowJobRecord {
@@ -276,8 +281,9 @@ export function extractDynamoRecord(payload: RelayPayload): OotWorkflowJobRecord
   const wf = cb.workflow;
   const pr = cb.payload?.pull_request;
 
-  const dynamoKey = `${trusted.verified_repo}/${cb.delivery_id}/${wf.name}`;
-  const now = new Date().toISOString();
+  const jobName = wf.job_name ?? "default";
+  const runAttempt = wf.run_attempt ?? 1;
+  const dynamoKey = `${trusted.verified_repo}/${cb.delivery_id}/${wf.name}/${jobName}/${runAttempt}`;
 
   const record: OotWorkflowJobRecord = {
     dynamoKey,
@@ -289,14 +295,31 @@ export function extractDynamoRecord(payload: RelayPayload): OotWorkflowJobRecord
     delivery_id: cb.delivery_id,
     workflow_run_url: wf.url ?? "",
     workflow_name: wf.name,
-    queue_time: trusted.ci_metrics?.queue_time,
-    execution_time: trusted.ci_metrics?.execution_time,
-    started_at: now,
+    job_name: jobName,
+    run_attempt: runAttempt,
   };
+
+  // Only set timing metrics when the relay provides a non-null value.
+  // in_progress sets queue_time; completed sets execution_time.
+  // Using UpdateItem ensures the completed callback doesn't clobber
+  // queue_time with null.
+  if (trusted.ci_metrics?.queue_time != null) {
+    record.queue_time = trusted.ci_metrics.queue_time;
+  }
+  if (trusted.ci_metrics?.execution_time != null) {
+    record.execution_time = trusted.ci_metrics.execution_time;
+  }
+
+  // Use downstream-reported timestamps, not HUD wall-clock time
+  if (wf.started_at) {
+    record.started_at = wf.started_at;
+  }
 
   if (wf.status === "completed") {
     record.conclusion = wf.conclusion ?? undefined;
-    record.completed_at = now;
+    if (wf.completed_at) {
+      record.completed_at = wf.completed_at;
+    }
     if (wf.test_results) {
       const tr = wf.test_results;
       if (typeof tr.total === "number") record.total_tests = tr.total;
@@ -313,52 +336,39 @@ export function extractDynamoRecord(payload: RelayPayload): OotWorkflowJobRecord
 }
 ```
 
-#### Payload Validation
+#### Schema Validation (at the Relay)
 
-The utility library validates the incoming relay payload structure:
+Schema validation is performed by the relay's Result Handler before forwarding to HUD. The relay validates required fields (`delivery_id`, `workflow.status`), status enum values, and conclusion presence on completed callbacks. HUD trusts that the relay has already validated the structure.
 
-```typescript
-export function validateRelayPayload(body: any): RelayPayload {
-  if (!body?.trusted?.verified_repo) {
-    throw new ApiError(400, "Missing trusted.verified_repo");
-  }
-  const cb = body?.untrusted?.callback_payload;
-  if (!cb) {
-    throw new ApiError(400, "Missing untrusted.callback_payload");
-  }
-  if (!cb.delivery_id) {
-    throw new ApiError(400, "Missing delivery_id");
-  }
-  if (!cb.workflow?.status) {
-    throw new ApiError(400, "Missing workflow.status");
-  }
-  if (cb.workflow.status !== "in_progress" && cb.workflow.status !== "completed") {
-    throw new ApiError(400, `Invalid workflow.status: must be "in_progress" or "completed".`);
-  }
-  if (cb.workflow.status === "completed" && !cb.workflow.conclusion) {
-    throw new ApiError(400, "workflow.conclusion is required when status is completed");
-  }
-  return body as RelayPayload;
-}
+The relay's validation logic (Python, in `result_handler.py`):
+
+```python
+try:
+    delivery_id = body["delivery_id"]
+    status = body["workflow"]["status"]
+except (KeyError, TypeError) as exc:
+    raise HTTPException(400, f"callback body missing required field: {exc}")
 ```
 
 #### Hop 3: HUD API → DynamoDB
 
-The HUD API writes the flattened workflow job record to DynamoDB:
+The HUD API writes the flattened workflow job record to DynamoDB using `UpdateItem` with `SET` expressions (not `PutItem`):
 
 ```
-dynamoKey = `${trusted.verified_repo}/${delivery_id}/${workflow_name}`
+dynamoKey = `${trusted.verified_repo}/${delivery_id}/${workflow_name}/${job_name}/${run_attempt}`
 ```
 
-- For "in_progress" callbacks: creates the record with minimal fields + `status: "in_progress"`
-- For "completed" callbacks: overwrites the record with full payload including `conclusion`, test counts, failed test details (as JSON string), artifact URL, and environment metadata
+- For "in_progress" callbacks: creates the record with initial fields (`status`, `started_at`, `queue_time`, etc.) via `UpdateItem`
+- For "completed" callbacks: merges `completed` fields (`conclusion`, `completed_at`, `execution_time`, test counts, artifact URL) into the existing record via `UpdateItem` — only non-null fields are set, so `queue_time` (set during `in_progress`) is preserved
+
+Using `UpdateItem` instead of `PutItem` ensures that the `completed` callback does not overwrite fields that were set during `in_progress` (e.g. `queue_time`, `started_at`) with `null` values.
 
 #### Hop 4: DynamoDB → ClickHouse (Automatic)
 
 This hop requires **no new code**. The existing infrastructure handles it:
 
 1. **DynamoDB Streams** (enabled on the table with `NEW_AND_OLD_IMAGES`) captures every write
-2. **`clickhouse-replicator-dynamo`** receives stream events and inserts `completed` records into ClickHouse (`in_progress` records are filtered out — they serve only as mutable state in DynamoDB)
+2. **`clickhouse-replicator-dynamo`** receives stream events and inserts **both** `in_progress` and `completed` records into ClickHouse. `SharedReplacingMergeTree` handles deduplication — when a `completed` record arrives for the same `dynamoKey`, it replaces the `in_progress` row after merges.
 3. **Replicator mapping** (1 new line): `"torchci-oot-workflow-job": "default.oot_workflow_job"` added to `SUPPORTED_TABLES` in `lambda_function.py`
 
 ### Storage Design
@@ -369,7 +379,7 @@ Table: `torchci-oot-workflow-job`
 
 | Field | Type | Description |
 |-------|------|-------------|
-| `dynamoKey` (hash key) | String | `{repo}/{delivery_id}/{workflow_name}` |
+| `dynamoKey` (hash key) | String | `{repo}/{delivery_id}/{workflow_name}/{job_name}/{run_attempt}` |
 | `status` | String | `in_progress` or `completed` |
 | `downstream_repo` | String | Downstream repo `org/name` (from `trusted.verified_repo`) |
 | `upstream_repo` | String | Upstream repo (e.g. `pytorch/pytorch`) |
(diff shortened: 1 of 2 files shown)


ZainRizvi (commented):
A few quick high level comments based on the PR description (I really appreciate the diagram btw!):

- I see some payload validation happening in HUD.  Why not have that in Relay Server instead? In HUD, auth check should likely be limited to authenticating the relay server's header value and a maaaaybe payload cap to ensure we don't overburden the database.  Everything else probably will find a better home in Relay Server, where the logic can be centralized. And the payload cap is likely worth adding to the relay server as well (HUD's check is more of a sanity check as the db guardian.)

- Why does the Storage box show that only completed jobs will be replicated to ClickHouse? I think the default behavior would be to replicate both in progress and completed, and that would offer more value as well. Current practice for in-tree CI is we update the job status in clickhouse (in progress -> completed) and we should do the same for OOT too. That's why we're storing the data in dynamo, else we could have stored it in S3

Security note: We should give the relay tool it's own unique header to send to hud and only that header should be allowed to make these relay api calls.  The X-Hud-Internal-Bot is a more widely used key and so far it has only been used for readonly operations. Let's not give it read-write access since the key is at higher risk of leaking.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,954 @@
...
+
+```typescript
+export function extractDynamoRecord(payload: RelayPayload): OotWorkflowJobRecord {
+  const { trusted, untrusted } = payload;
+  const cb = untrusted.callback_payload;
+  const wf = cb.workflow;
+  const pr = cb.payload?.pull_request;
+
+  const dynamoKey = `${trusted.verified_repo}/${cb.delivery_id}/${wf.name}`;
+  const now = new Date().toISOString();
```
ZainRizvi: There are a couple of bugs here:

- `started_at` is set to `now()` on every callback, including `completed`. That's the time HUD received the POST, not when the job started. On a completed callback, `started_at ≈ completed_at`, so the ClickHouse `duration_seconds` alias will always be ≈ 0. Neither `started_at` nor `completed_at` are available in the `github` context, so the action should capture its own wall-clock time and include it in the payload — `in_progress` sends `started_at`, `completed` sends `completed_at`. Then HUD stores what the action reports instead of using `now()`.

- `queue_time` is set from `trusted.ci_metrics?.queue_time`, which is `null` on completed callbacks (the relay only sends it on `in_progress`). This needs to skip setting `queue_time` when it's null rather than writing null over a valid value. Same applies to `execution_time` on `in_progress` callbacks.

Also, the DynamoDB write needs to use `UpdateItem` with `SET` expressions instead of `PutItem`. PutItem does a full item replace, so even with the above fixes, the completed callback would clobber `queue_time` and `started_at` from the in_progress record. `UpdateItem` lets each callback write only its own fields.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,954 @@
...
+The `extractDynamoRecord` function flattens this into a single record:
+
+```typescript
+export function extractDynamoRecord(payload: RelayPayload): OotWorkflowJobRecord {
+  const { trusted, untrusted } = payload;
+  const cb = untrusted.callback_payload;
+  const wf = cb.workflow;
+  const pr = cb.payload?.pull_request;
+
+  const dynamoKey = `${trusted.verified_repo}/${cb.delivery_id}/${wf.name}`;
```
ZainRizvi: This key needs to include the job name and attempt number. The callback action runs inside a specific job, and if a downstream repo uses it in multiple jobs within the same workflow (build, test-cpu, test-gpu), all those jobs share the same `github.workflow` name and would write to the same dynamoKey — last writer wins, the rest are silently lost. And without the attempt number, job retries would overwrite the original run.

Let's add `github.job` and `github.run_attempt` to the key: `{repo}/{delivery_id}/{workflow_name}/{job_name}/{attempt}`. That gives each job its own row, preserves retry history, and makes the `/oot/[org]/[repo]` matrix view ("columns = OOT jobs") actually useful.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,954 @@
...
+    pr_number = {pr: UInt64}
+ORDER BY
+    downstream_repo, started_at DESC
+```
+
+The `OotPrSection` React component is integrated into the existing PR page (`torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx`) below the existing `CommitInfo` section.
+
+### Sample Payloads
+
+#### In-Progress Callback
```
ZainRizvi: These samples don't match the actual wire format. They show a flat structure with top-level `status`, `head_sha`, `pr_number`, etc. But the L2 PR's `action.yml` sends a nested structure — `delivery_id` and `payload` from the dispatch envelope, with a `workflow` sibling containing `status`/`conclusion`/`name`/`url`. Your own TypeScript code (`RelayPayload`, `extractDynamoRecord`) parses the nested format correctly, so the code is right — but a downstream developer implementing from these samples would produce payloads the system can't parse. These need to match the actual `{delivery_id, payload, workflow}` structure from the action.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,954 @@
...
+
+- For "in_progress" callbacks: creates the record with minimal fields + `status: "in_progress"`
+- For "completed" callbacks: overwrites the record with full payload including `conclusion`, test counts, failed test details (as JSON string), artifact URL, and environment metadata
+
+#### Hop 4: DynamoDB → ClickHouse (Automatic)
+
+This hop requires **no new code**. The existing infrastructure handles it:
+
+1. **DynamoDB Streams** (enabled on the table with `NEW_AND_OLD_IMAGES`) captures every write
+2. **`clickhouse-replicator-dynamo`** receives stream events and inserts `completed` records into ClickHouse (`in_progress` records are filtered out — they serve only as mutable state in DynamoDB)
```
ZainRizvi: `in_progress` records need to be replicated to ClickHouse too, not just `completed`. Current practice for in-tree CI is we update job status in ClickHouse (in_progress → completed) and OOT should follow the same pattern. `SharedReplacingMergeTree` already supports upserts. Without this, the `/pr/N` page wouldn't be able to show running jobs.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,954 @@
...
+
+#### Hop 2: Result Handler → HUD API
+
+| Step | Action | Failure Response |
+|------|--------|------------------|
+| 1 | Validate `X-Hud-Internal-Bot` header | `401 Unauthorized` |
+| 2 | Schema validation + payload caps (2MB max body) | `400 Bad Request` |
+| 3 | Write to DynamoDB | `502 Bad Gateway` |
+
+The HUD API endpoint (`torchci/pages/api/oot/results.ts`) follows the existing `webhookToDynamo` pattern:
```
ZainRizvi, line 173: Schema validation and field extraction should live in the Relay Server, not HUD. The relay already does OIDC, allowlist checks, and rate limiting — adding schema validation there keeps HUD simple (authenticate the relay, write what it's told) and avoids duplicating validation if other consumers of relay data appear later. HUD should still enforce a payload size cap as a safety net.

ZainRizvi (commented):
Nice work — the architecture diagrams are really clear and the RFC builds on RFC-0050 well. A few things to sort out, mostly around aligning with the L2 PR's actual wire format and a couple of bugs in the extraction logic.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,954 @@
...
+| Max `failed_tests_json` entries | 1,000 per request | HUD API schema validation |
+| Max `stacktrace` length | 4 KB per test | HUD API (truncate before write) |
+| Max `message` length | 1 KB per test | HUD API (truncate before write) |
+
+### Authentication Flow
+
+| Hop | From → To | Auth Mechanism | Credential Scope |
+|-----|-----------|----------------|------------------|
+| 1 | Downstream → Result Handler | OIDC token (GHA-issued, 5 min TTL) | No secrets needed by downstream |
+| 2 | Result Handler → HUD API | `X-Hud-Internal-Bot` header | Same pattern as DrCI / trymerge |
```
ZainRizvi: Let's use a dedicated auth header for the relay instead of `X-Hud-Internal-Bot`. That header is shared across multiple systems (DrCI, trymerge) and has only been used for read-only operations. Since this is a write path, giving the relay its own key limits blast radius if any single one leaks.

subinz1:
> * I see some payload validation happening in HUD.  Why not have that in Relay Server instead? 

Hello @ZainRizvi, Agreed — moved schema validation to the relay (Hop 1, step 5). The relay now validates delivery_id, workflow.status, and conclusion presence before forwarding. HUD is simplified to: auth (X-OOT-Relay-Token) + payload cap (2MB) + UpdateItem write. This aligns with the L2 PR's result_handler.py which already validates these fields at the relay level

subinz1:
> * Why does the Storage box show that only completed jobs will be replicated to ClickHouse? I think the default behavior would be to replicate both in progress and completed, and that would offer more value as well. Current practice for in-tree CI is we update the job status in clickhouse (in progress -> completed) and we should do the same for OOT too. That's why we're storing the data in dynamo, else we could have stored it in S3

Fixed — removed all "completed only" language. Both in_progress and completed records now replicate to ClickHouse. SharedReplacingMergeTree handles the deduplication — when the completed record arrives for the same dynamoKey, it replaces the in_progress row after merges. This matches in-tree CI behavior and allows the /pr/N page to show running jobs.

> Security note: We should give the relay tool it's own unique header to send to hud and only that header should be allowed to make these relay api calls. The X-Hud-Internal-Bot is a more widely used key and so far it has only been used for readonly operations. Let's not give it read-write access since the key is at higher risk of leaking.

Changed everywhere —  diagrams, data flow table, authentication flow table, code samples, and implementation plan. The relay now uses a dedicated X-OOT-Relay-Token header instead of the shared X-Hud-Internal-Bot. This isolates the OOT write path so rotating/revoking the token doesn't affect DrCI/trymerge.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: All three fixed:

* started_at/completed_at now use downstream-reported timestamps (wf.started_at, wf.completed_at) instead of new Date().toISOString(). Note: these fields need to be added to the L2 PR's callback action (action.yml), which currently doesn't send wall-clock timestamps.
* queue_time/execution_time are only set when the relay provides a non-null value. in_progress sets queue_time; completed sets execution_time. Neither overwrites the other.
* Changed from PutItem to UpdateItem with SET expressions throughout — Hop 3, data flow table, code samples, and implementation plan.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Thanks you for the thorough review @ZainRizvi, Changed from {repo}/{delivery_id}/{workflow_name} to {repo}/{delivery_id}/{workflow_name}/{job_name}/{run_attempt}. Added job_name and run_attempt to both DynamoDB and ClickHouse schemas. The L2 callback action needs to be updated to include github.job and github.run_attempt in the payload.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Replaced all flat-structure samples with the actual nested wire format, shown at two stages:

Stage 1 (downstream → relay): {event_type, delivery_id, payload, workflow} — matching the callback action's output from action.yml
Stage 2 (relay → HUD): {trusted: {verified_repo, ci_metrics}, untrusted: {callback_payload: ...}} — matching forward_to_hud in hud.py
Also added a note about fields (job_name, run_attempt, started_at, completed_at) that need to be added to the L2 action.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Fixed — both in_progress and completed records now replicate to ClickHouse. The replicator forwards all DynamoDB stream events without filtering. SharedReplacingMergeTree handles the upsert — completed replaces in_progress for the same dynamoKey after merges.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: We have moved — schema validation is now at the relay (Hop 1, step 5). The HUD API endpoint is simplified to: validate X-OOT-Relay-Token header, enforce 2MB payload cap, extract/flatten the record, and UpdateItem write. HUD trusts that the relay has already validated the structure.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Definitely @ZainRizvi, 
we have changed to X-OOT-Relay-Token — a dedicated header scoped only to the OOT relay write path. Added an explanation in the RFC that this isolates the write path so key rotation or revocation doesn't affect other internal HUD consumers (DrCI, trymerge).

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,1045 @@
...
+> The callback token is not strictly required for L2 (dashboard only). However, L1 is where the token gets minted, and once L1 is deployed and downstream repos depend on the current `client_payload` shape, retrofitting it becomes a breaking change. We recommend adding it now as optional groundwork for L3/L4.
+
+#### State Machine for Status Transitions
+
+To prevent rollback and replay attacks at the relay level, we propose a Redis-backed state machine for status transitions:
+
+| Transition | Action |
+|------------|--------|
+| No record → `in_progress` | Accepted |
+| No record → `completed` | Accepted (handles missed `in_progress` callback) |
```
KarhouTam: Why No record → `completed` should be accepted?

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,1045 @@
...
+    "repository": { "full_name": "pytorch/pytorch" }
+  },
+  "workflow": {
+    "status": "in_progress",
+    "conclusion": null,
+    "name": "xpu-ci",
+    "url": "https://github.com/{org}/{repo}/actions/runs/24033272679",
+    "job_name": "test-xpu-float32",
+    "run_attempt": 1,
+    "started_at": "2025-04-28T10:15:30Z"
```
KarhouTam: In my state machine implementation, the Redis key now includes `check_run_id` instead of `run_attempt`. Since both values change with reruns, I believe `check_run_id` is a better fit for representing a unique job than `{job_name}:{run_attempt}`. Accordingly, the Redis state key has been updated to `oot:state:{delivery_id}:{repo}:{check_run_id}`. `run_attempt` still in the dict, but not used by relay server.

To improve HUD usability, I added `run_id` and `job_name` to the `workflow` dictionary, which keep consistent among all runs. This allows the HUD to use a grouping key like `{job_name}:{run_id}`, preventing a single downstream workflow from appearing as multiple rows due to reruns. Under this approach, every unique `check_run_id` is treated as a separate job.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,1045 @@
...
+| Fabrication blocking | Identity check only | Token locks to specific PR/SHA |
+| Orphaned job detection | No way to know | Unconsumed Redis key = orphaned |
+| Queue time (trusted) | Redis lookup, can miss | Embedded in token, tamper-proof |
+| Per-dispatch rate limiting | Per-repo only | Exactly 2 callbacks per dispatch |
+| L3/L4 merge gating | Unsafe (self-reported) | Cryptographic proof of dispatch |
+
+> [!NOTE]
+> The callback token is not strictly required for L2 (dashboard only). However, L1 is where the token gets minted, and once L1 is deployed and downstream repos depend on the current `client_payload` shape, retrofitting it becomes a breaking change. We recommend adding it now as optional groundwork for L3/L4.
+
+#### State Machine for Status Transitions
```
KarhouTam, line 603: The current state machine differs slightly from this RFC. Here are the state transitions:

According to previous comments in the L2 implementation PR, this state machine has three states:
- `DISPATCHED` (set only once at the webhook handler per repository; this state is repo-level and every job needs it to update its state).
- `IN_PROGRESS` (job-level, cannot be duplicated; reruns are treated as separate jobs).
- `COMPLETED` (job-level, cannot be duplicated; reruns are treated as separate jobs).

Note that the direction graph below is for a single check run, reruns have different `check_run_id` and are treated as separate jobs, so they won't violate the state machine since they won't have a prior `IN_PROGRESS` or `COMPLETED` record.

```mermaid
stateDiagram-v2
    direction LR
    [*] --> DISPATCHED: webhook sends
    DISPATCHED --> IN_PROGRESS: first callback
    IN_PROGRESS --> COMPLETED: completion
    IN_PROGRESS --> IN_PROGRESS: ❌ duplicate
    DISPATCHED --> COMPLETED: ❌ skip IN_PROGRESS
    COMPLETED --> COMPLETED: ❌ duplicate
    COMPLETED --> IN_PROGRESS: ❌ wrong direction
    [*] --> IN_PROGRESS: ❌ no dispatch
    [*] --> COMPLETED: ❌ no dispatch
```

**State types:**
- `DISPATCHED`: Repo-level state (`check_run_id=DISPATCH_CHECK_RUN_ID`) — one per repository.
- `IN_PROGRESS` / `COMPLETED`: Job-level states — independent per job.

**Valid flows:**
- Main: `DISPATCHED → IN_PROGRESS → COMPLETED` (linear)
- Reruns: `IN_PROGRESS → IN_PROGRESS` or `COMPLETED → IN_PROGRESS` (update timestamps)

**Rejected:**
- Skipping `IN_PROGRESS`, duplicate `COMPLETED`, or missing `DISPATCHED` state.
- Replay attack: duplicate `IN_PROGRESS` (same `check_run_id`).

KarhouTam (commented):
Hey, @jewelkm89. Thanks for this RFC, really a great work! According to the previous comments and suggestions on CRCR L2 implementation [PR](https://github.com/pytorch/test-infra/pull/7967), some inconsistencies existed between the current implementation and this RFC. I've left some comments. PTAL!

cc @fffrog @can-gaa-hou

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,1045 @@
...
+          skipped?: number;
+          failures?: any[];
+        };
+      };
+    };
+  };
+}
+```
+
+The `extractDynamoRecord` function flattens this into a single record. It uses `started_at` and `completed_at` timestamps reported by the downstream action (wall-clock time when the callback was sent), rather than generating timestamps at HUD. Timing metrics (`queue_time`, `execution_time`) are only set when the relay provides a non-null value, preventing `completed` callbacks from clobbering `queue_time` with `null`:
```
can-gaa-hou, line 280: For the retry scenario, `queue_time` may be an invalid value. I suppose we should not update this value in this case.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,1045 @@
...
+| 1 | Verify OIDC token signature against GitHub JWKS | `401 Unauthorized` |
+| 2 | Check `repository` claim against cached allowlist | `403 Forbidden` |
+| 3 | Verify repo is authorized at L2 or above | `403 Forbidden` |
+| 4 | Per-repo rate limit check (20 req/min default) | `429 Too Many Requests` |
+| 5 | Validate required fields: `delivery_id`, `workflow.status` | `400 Bad Request` |
+| 6 | Forward validated payload to HUD API | — |
+
+The relay's Result Handler validates the callback body and produces a `{trusted, untrusted}` payload:
+
+- `trusted` — relay-generated fields: `verified_repo` (OIDC-proven identity) and `ci_metrics` (relay-measured `queue_time`, `execution_time`)
```
KarhouTam: `downstream_repo_level` is included for HUD convenience now.

Values: `L1` ~ `L4` (string)

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,1045 @@
...
+      };
+      workflow: {
+        status: string;
+        conclusion?: string | null;
+        name: string;
+        url: string;
+        job_name?: string;
+        run_attempt?: number;
+        started_at?: string;
+        completed_at?: string;
```
KarhouTam, line 267: `workflow` now is like:
```python
workflow: dict = {
    "schema_version": ...,
    "status": ...,
    "conclusion": ...,
    "name": ...,
    "url": ...,
    "run_attempt": ...,
    "job_name": ...,
    "check_run_id": ...,
    "run_id": ...,
    "started_at": ...,
    "completed_at": ...,
    "test-results": ...
}
```

KarhouTam (commented):
More inconsistencies found.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Hello @KarhouTam,
During the initial design, we have decided that if the `in_progress` couldn't make it to the DynamoDB due to outages, the downstream repo should still be allowed to insert the `completed` record. Please refer [comment](https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?disco=AAAB3xDqDYg).

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Thanks @KarhouTam. Good call — updated the RFC and reference implementation to source downstream_repo_level from trusted (relay-determined from the allowlist) instead of from the untrusted callback payload. Changes:

RFC:

- Added `downstream_repo_level` to the `trusted` fields description and `RelayPayload` interface
- `extractDynamoRecord` now reads it from `trusted.downstream_repo_level`
- Both sample payloads (`in_progress` and `completed`) now include `"downstream_repo_level": "L2"` in the `trusted` section

Reference implementation (`subinz1/test-infra#1`):

- Added `downstream_repo_level` to `RelayTrusted` and `OotWorkflowJobRecord` interfaces
- `extractDynamoRecord` extracts it from `trusted.downstream_repo_level`

The DynamoDB and ClickHouse schemas already had the column from a prior commit — only the ingestion/extraction path was missing.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Thanks for the updated field list @KarhouTam. Updated the RFC and reference implementation to align:

- Added `schema_version`, `check_run_id`, and `run_id` to the workflow interface, DynamoDB/ClickHouse schemas, queries, and sample payloads
- Switched `dynamoKey` from `run_attempt` to `check_run_id` for per-execution uniqueness
- Fixed `test-results` key to use the hyphenated form matching the L2 action (was `test_results` — would have silently failed in JS)

One question: should `test-results` stay hyphenated, or would it be cleaner to normalize to `test_results` in the action? Hyphenated keys require bracket notation in TS (`wf["test-results"]`). Happy either way — just want to confirm the convention.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Good catch @can-gaa-hou. On retries (`run_attempt` > 1), `queue_time = dispatch_timestamp → rerun_in_progress_timestamp` could be hours or days old — meaningless as a CI metric.

The fix belongs on the relay side: the relay should set `queue_time = null` when `run_attempt > 1`. On the HUD side, we already only write `queue_time` when the relay provides a non-null value, so no HUD changes are needed.

Updated the RFC to document this behavior.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Great improvements @KarhouTam. Aligned the RFC with your changes:

- **State machine key**: Updated from `run_id` to `check_run_id` — `oot:state:{delivery_id}:{repo}:{check_run_id}`. Since `check_run_id` is GitHub-assigned and unique per job execution, it's a better uniqueness key than `{job_name}:{run_attempt}.`
- **3-state mode**l: Adopted the `DISPATCHED → IN_PROGRESS → COMPLETED` flow with no shortcuts. `DISPATCHED` is repo-level (set by webhook handler), `IN_PROGRESS/COMPLETED` are job-level per `check_run_id`. Added your mermaid diagram.
- **New fields**: `check_run_id`, `run_id`, `schema_version` added to the workflow interface, DynamoDB/ClickHouse schemas, sample payloads, and the reference implementation.
- **DynamoKey**: Switched from` {job_name}/{run_attempt}` to `{job_name}/{check_run_id}` for per-execution uniqueness.
- **HUD grouping**: Frontend groups by `job_name` and keeps the latest `run_attempt` per PR, preventing reruns from appearing as duplicate rows.

run_attempt is still in the dict and stored for informational display, but not used for keying.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: After further review and aligning with the L2 implementation, we've updated the RFC to reject `No record → completed`. With `check_run_id`-based state tracking and the `DISPATCHED` prerequisite, accepting a completed without prior `in_progress` would bypass the state integrity check. The strict `DISPATCHED → IN_PROGRESS → COMPLETED` flow provides better replay and fabrication protection.

KarhouTam: Currently, when an `in_progress` callback is lost, the system returns a 400 error, which requires the on-call operator to manually rerun the job. While this behavior was designed to reflect real-world network instability, we have decided after thoughtful internal discussion that it would be more efficient to ensure a `completed` callback is always preceded by an `in_progress` callback.

cc @fffrog @can-gaa-hou

subinz1: Thanks for confirming @KarhouTam. This aligns with the update we made to the RFC — `No record → completed` is now rejected, and the strict `DISPATCHED → IN_PROGRESS → COMPLETED` flow is enforced. If `in_progress` is lost, the operator reruns the job rather than accepting a completed without state integrity. The RFC reflects this as the agreed-upon behavior.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
KarhouTam: Good one! Actually, in the latest `action.yml`, we changed it to `workflow["test_results"]` already.

https://github.com/KarhouTam/test-infra/blob/c2833bd9eee6f7299311695597ad1e57ad974c7b/.github/actions/cross-repo-ci-relay-callback/action.yml#L121-L126

Sorry for the misleading comment.

subinz1: Thanks @KarhouTam. We've updated both the RFC and reference implementation to use `test_results` (underscore) — it matches the L2 action's Python code (`workflow["test_results"] = ...`) and is more convenient for TypeScript (dot notation vs bracket notation). Also aligned `test_results` to be summary counts only (total/passed/failed/skipped), with detailed test results going via `artifact_url` as the action describes.

jewelkm89:
Hello @albanD, @huydhn and @atalman
Request your review for the RFC as well.
Associated PR is https://github.com/pytorch/test-infra/pull/8069

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,1078 @@
...
+
+1. **Result Handler** (existing, from relay): receives callbacks from downstream CI, verifies OIDC, checks allowlist, forwards to HUD
+2. **HUD API** (`/api/oot/results`): validates auth (`x-hud-internal-bot`), enforces payload caps, writes to DynamoDB via `UpdateItem`
+3. **DynamoDB** (`torchci-oot-workflow-job`): mutable storage supporting the two-callback model (`in_progress` → `completed`)
+4. **ClickHouse** (`default.oot_workflow_job`): analytical storage, replicated automatically via the existing `clickhouse-replicator-dynamo` replicator
+
+**Read path:**
+
+1. **Global CRCR Summary** (`/crcr`): cross-repo health overview for CI maintainers
+2. **Per-Backend Dashboard** (`/crcr/[org]/[repo]`): detailed job grid for a single downstream repo
```
atalman, line 54: Lets also fix sync this change to rfc: https://github.com/pytorch/rfcs/blob/master/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md I belive it needs to be updated to use `` /oot/[org]/[repo] ``

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,1078 @@
...
+
+#### Rate Limiting
+
+**At the relay (per-repo)**
+
+```
+Key:    oot:rate:{verified_repo}
+Value:  counter (atomic INCR)
+TTL:    1 minute sliding window
+Limit:  20 requests/minute per repo (configurable via RATE_LIMIT_PER_MIN)
```
atalman: This should be 60 now I believe

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Good catch @atalman — fixed in d207b51. Updated RFC-0050 to use `/crcr/[org]/[repo]` and `/crcr` (the routes were renamed from `/oot` to `/crcr` in pytorch/test-infra#8142). Also added the explicit global summary route `hud.pytorch.org/crcr` that was previously missing from RFC-0050's HUD Integration section.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
subinz1: Updated in 1bd5dab — changed from 20 to 60 requests/minute per repo to reflect the current deployed Lambda config.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
```diff
@@ -0,0 +1,1078 @@
...
+
+4. ~~**Backfill on failure**~~ **Resolved**: The callback handler retries HUD 5xx/network errors with exponential backoff (up to 3 retries). HUD 4xx errors propagate immediately. No dead-letter queue for now.
+
+5. **Callback token adoption timeline**: the signed callback token strengthens security significantly but requires coordination with L1 deployment. Can be added incrementally.
+
+6. ~~**Silent HUD forward skip**~~ **Resolved**: `HUD_API_URL` is now provisioned as a Lambda default ([#8147](https://github.com/pytorch/test-infra/pull/8147)). The relay logs and returns `"skipped"` when the URL is not set (local dev only).
+
+7. **`timed_out`/`cancelled` conclusion support**: The composite action (`cross-repo-ci-relay-callback`) only accepts `success` or `failure` as valid `conclusion` values when `status=completed`. Clarification of accepted values is being addressed in [#8173](https://github.com/pytorch/test-infra/pull/8173).
+
+## Implementation Plan
```
albanD, line 999: Note that the description of the specific PRs is not very accurate. Should we just remove it? As it only really matters what the end result is, not the intermediate states

albanD (approved):
Besides the innacurate PR descriptions, that sounds fine.
I would remove them tbh as I don't think they add much.

RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md
groenenboomj: I agree, this information isn't needed. We can cull it.

atalman (approved):
lgtm

jewelkm89:
Thank you @albanD  for the review. Kindly help merge the RFC as well
cc @atalman

groenenboomj:
With the last change I think we have addressed all feedback. Ready to roll.

> Thank you @albanD for the review. Kindly help merge the RFC as well cc @atalman

PR #109, opened 2026-09-10
Add RFC-0060: PyTorch install layouts

zklaus:
PyTorch is installed in one of three shapes (artifacts beside the Python
sources, split from them, or without them), and the documentation describes
the result of only one of the six commands that produce them. Code that
derives asset paths from __file__ is right in some shapes and wrong in
others, which is how seven call sites broke under editable installs after
the scikit-build-core migration.

The RFC documents the shapes from listings of the shipped 2.14.0 wheels and
libtorch zips and of real editable installs on Linux, macOS and Windows;
states two resolution rules for in-tree code (resources through
importlib.resources behind a thin helper, directories for external tools
through the existing location helper, both anchored on the import system);
proposes one documentation page linked from CONTRIBUTING.md and the C++
docs; and aligns the standalone cmake --install recipe and
tools/build_libtorch.py with the zips through three build defaults.

Opened as a draft for discussion. Authored with assistance from Claude Code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RFC-0060-install-layouts.md
```diff
@@ -0,0 +1,506 @@
...
+
+Wheel layout also leaks into unrelated docs: `torch.compiler_aot_inductor.md`
+hardcodes `/path/to/python/install/site-packages/torch/share/cmake`, wrong
+for an editable install.
+
+Every failure is a conflation of layouts, so the layouts want documenting
+together, with the resolution rule stated beside them.
+
+
+## **Proposed Implementation**
```
atalman: Hi @zklaus 

Proposed Implementation" is mostly *description* — the three shapes, the inventory table, where each directory comes from — and the actual changes are scattered through prose. A reviewer can't tell what they're approving without assembling the list themselves, and whoever implements it has to re-derive the scope.

For a document whose thesis is "nobody wrote the rule down, so it broke twice," that's worth fixing. Here's the list I extracted; I'd put something like it at the top of Proposed Implementation, ahead of the descriptive material:

**Code**

1. Add `torch._utils_internal.resource(*parts)` — wraps `importlib.resources`, returns a `Traversable`.
2. Re-anchor `torch._utils_internal.get_file_path()` on `torch._C`'s spec instead of `__file__`.
3. Fix the known call sites: the seven from #195887, plus `torch.mps.compile_shader`, the inductor `codegen/aoti_runtime/` sources, `kernel/flex/templates/`, `kernel/templates/`, `csrc/inductor/aoti_runtime/model.h`, `torch.utils.model_dump`, `torch._export.serde.schema_check`.
4. Re-anchor `torchgen.gen.get_torchgen_root()` and add a torchgen resource entry point for `packaged/`.

**Build**

5. `docs/libtorch.rst` recipe adds `BUILD_PYTHON=OFF` and `BUILD_TEST=OFF`.
6. CMake emits `message(STATUS)` when `BUILD_PYTHON` is on outside scikit-build-core.
7. `cmake/ProtoBuf.cmake` sets `protobuf_INSTALL=OFF`.
8. `tools/build_libtorch.py` default prefix moves from `<checkout>/torch` to `<cwd>/libtorch`, gains `--install-prefix`, refuses an in-source prefix.
9. `.ci/pytorch/windows/arm64/build_libtorch.bat` adapted to read the new prefix.

**CI**

10. Editable-install job (#195888).
11. Standalone-layout job asserting the resulting tree.
12. Lint: no `__file__`-derived paths, no direct `importlib.resources`, outside the two helpers.

**Docs**

13. "Install layouts" page — three shapes, table, two rules.
14. Links from `CONTRIBUTING.md`, `docs/libtorch.rst`, `docs/cpp/source/installing.md`.
15. "Where things end up" paragraph in `CONTRIBUTING.md`.

Details list:
Generated with [Claude Code](https://claude.com/claude-code)

PR #111, opened 2026-09-18
RFC-00xx: Enable Distributed Cross Node Testing

skpark-rh:
# Summary
PyTorch currently only has distributed tests for single-node testing. This RFC aims to address this limitation and turn on cross node testing.  The solution is to add a no-op attribute, `local_rank`, to `MultiProcessTestBase` that differentiate what is global and what is local.

# Scope
Phase 1 is to add the attribute `local_rank` to `MultiProcessTestCase`.
Phase 2 is to change every instance of `torch.device` in the child classes of `MultiProcessTestCase` to refer to `self.local_rank` instead of `self.rank`.

# Feedback
In this RFC, I would like feed back on two questions.
1. Attribute name (`local_rank` vs alternatives), and whether it's set via env (LOCAL_RANK) or from computation.
2. Enabling Cross-node testing CI uptream?


skpark-rh:
I don't know what number this RFC should be so I left it `RFC-00xx`

